PR #14, opened 2024-08-21
fix use multi gpu device, sm metric is 0

lengrongfu:
Fixes: https://github.com/Project-HAMi/HAMi/issues/445

Test success:
image


lengrongfu:
@archlitchi PTAL

PR #20, opened 2024-09-03
add extern to make compile pass

A-transformer:


A-transformer:
verify it on T4 and A4000

archlitchi:
My deepest thanks, that really works!

PR #64, opened 2025-04-11
Fix: lock file loop and fd leak

dongjiang1989:
ref: https://github.com/Project-HAMi/HAMi-core/issues/62


dongjiang1989:
@archlitchi PTAL

PR #65, opened 2025-04-11
fix  return func by cuda version not always newest

huanghuangzym:
fix https://github.com/Project-HAMi/HAMi-core/issues/57


huanghuangzym:
vllm  cuda runtime 11.8  can not use some _v2 func in cuda driver 12.x。  
maybe some other cuda app has same problem,  when app run in different gpu node, may  support max cuda 11, or max cuda 12,we can promise it can run

PR #66, opened 2025-04-14
add build pipeline

huanghuangzym:


.github/workflows/build-src.yml
```diff
@@ -0,0 +1,26 @@
...
+    timeout-minutes: 1440
+    strategy:
+      fail-fast: true
+    name: Build libvgpu
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
```
dongjiang1989, line 20:
```suggestion
      -
        name: Set up QEMU
        uses: docker/setup-qemu-action@v3 
      - name: Set up Docker Buildx
        id: buildx
        uses: docker/setup-buildx-action@v3
```

dongjiang1989: Add linux/amd64,linux/arm64 muti-arch

huanghuangzym: purpose to add this only for check if pr has any build error?   in hami  repo  will include tihis libvgpu.so compile

archlitchi (approved):
/lgtm

PR #73, opened 2025-04-28
feat: Add cuCtxCreate_v3 context func hook

dongjiang1989:
cudaVersion >=11050

### DIFF
--- a/src/cuda/context.c
+++ b/src/cuda/context.c
@@ -45,6 +45,12 @@ CUresult cuCtxCreate_v2 ( CUcontext* pctx, unsigned int  flags, CUdevice dev ){
     return res;
 }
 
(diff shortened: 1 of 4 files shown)


dongjiang1989:
@archlitchi PTAL

dongjiang1989:
> do you have any issue related to this PR?

@archlitchi Thanks
When creating a client cuda ctx via `cuCtxCreate_v3` and specifying the execution affinity `CUexecAffinityParam` can limit the number of SMs that the ctx can use. 
But `cuCtxCreate_v3` is not hooked by hami-core. Hmmm....

archlitchi:
> > do you have any issue related to this PR?
> 
> @archlitchi Thanks When creating a client cuda ctx via `cuCtxCreate_v3` and specifying the execution affinity `CUexecAffinityParam` can limit the number of SMs that the ctx can use. But `cuCtxCreate_v3` is not hooked by hami-core. Hmmm....

wow, that's great, i'll try testing it on my environment first

PR #78, opened 2025-06-04
fix the display of memory in nvmlDeviceGetMemoryInfo v2

NUABO:
nvmlDeviceGetMemoryInfo v2 free memory is not displayed correctly


NUABO:
> do you have a related-issue describing what went wrong before your fixing? a bug screenshot is most appreciated

@archlitchi sorry for not having relevant screenshots. I looked at the code and found that ((nvmlMemory_v2_t *)memory)->used was assigned twice in the past process, and there was no free assignment. I think there is something wrong with this code.

PR #84, opened 2025-06-19
fix primary context memory usage to prevent GPU memory allocation failed

NUABO:
Update GPU memory usage after release the primary context memory to prevent oom_check failures and memory allocation failed


archlitchi:
nice fix, thanks

PR #97, opened 2025-07-10
[Bugfix] Fix the issue of duplicate statistics in  under multi-process

chaunceyjiang:
Before
image
image
image

It can be observed that when a user runs multiple processes in a pod and all of them use the GPU, a metric accumulation issue occurs.
Although we lock `shrreg`, after `unlock_shrreg`, another process immediately updates the metrics.

After
image


archlitchi (approved):
/lgtm

PR #117, opened 2025-08-13
build: optimize Dockerfile for best-effort image size reduction and cache usage

mayooot:


mayooot:
@archlitchi Could you please review this?

PR #124, opened 2025-08-28
fix: check memory null

DSFans2014:
if memory is NULL, the program may coredump at https://github.com/Project-HAMi/HAMi-core/blob/main/src/nvml/hook.c#L325


archlitchi (approved):
/lgtm

PR #129, opened 2025-09-22
fix: avoid pthread_self truncation in dlsym recursion guard

xrwang8:
### Summary

  - stores the thread id in the dlsym recursion guard as pthread_t instead of int and compares via pthread_equal
  - adds the missing <pthread.h> include required for the new type

  This removes the undefined behaviour where pthread_self() could be truncated on 64‑bit glibc systems, which previously allowed the recursion guard to misidentify threads and potentially skip or mis-
  handle nested dlsym calls.

### DIFF
--- a/src/libvgpu.c
+++ b/src/libvgpu.c
@@ -1,6 +1,7 @@
 //#include "memory_limit.h"
 #include <fcntl.h>
 #include <dlfcn.h>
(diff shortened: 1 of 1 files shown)


archlitchi (approved):
/lgtm

PR #130, opened 2025-09-22
fix: change return type of nvml_to_cuda_map to int for consistency

xrwang8:
### Summary

  - change nvml_to_cuda_map to return int so it can safely signal failure with -1
  - update the function declaration and callers to rely on the signed return value

  This removes the implementation-defined behavior where an unsigned -1 could be interpreted as a huge positive index, bypassing the <0 guard and leading to out-of-bounds access when NVML devices are not
  visible to CUDA (e.g., under CUDA_VISIBLE_DEVICES remapping).

### DIFF
--- a/src/multiprocess/multiprocess_utilization_watcher.c
+++ b/src/multiprocess/multiprocess_utilization_watcher.c
@@ -105,7 +105,7 @@ int nvml_to_cuda_map(unsigned int nvmldev){
     return -1;
 }
 
-unsigned int cuda_to_nvml_map(unsigned int cudadev){
+int cuda_to_nvml_map(unsigned int cudadev){
     return cuda_to_nvml_map_array[cudadev];
 }
 
--- a/src/multiprocess/multiprocess_utilization_watcher.c
+++ b/src/multiprocess/multiprocess_utilization_watcher.c
@@ -95,13 +95,12 @@ long delta(int up_limit, int user_current, long share) {
   return share;
 }
 
-unsigned int nvml_to_cuda_map(unsigned int nvmldev){
+int nvml_to_cuda_map(unsigned int nvmldev){
     unsigned int devcount;
     CHECK_NVML_API(nvmlDeviceGetCount_v2(&devcount));
-    int i=0;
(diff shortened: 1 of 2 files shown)


src/multiprocess/multiprocess_utilization_watcher.c
```diff
@@ -95,18 +95,17 @@ long delta(int up_limit, int user_current, long share) {
...
 }
 
-unsigned int nvml_to_cuda_map(unsigned int nvmldev){
+int nvml_to_cuda_map(unsigned int nvmldev){
     unsigned int devcount;
     CHECK_NVML_API(nvmlDeviceGetCount_v2(&devcount));
-    int i=0;
-    for (i=0;i<devcount;i++){
+    for (unsigned int i = 0; i < devcount; i++){
         if (cuda_to_nvml_map(i)==nvmldev)
```
DSFans2014, line 102: compared between signed and unsigned integer. it's better to convert to the same type?

xrwang8: This removes the implementation-defined behavior where an unsigned -1 could be interpreted as a huge positive index, bypassing the <0 guard and leading to out-of-bounds access when NVML devices are not
visible to CUDA (e.g., under CUDA_VISIBLE_DEVICES remapping)

xrwang8: This removes the implementation-defined behavior where an unsigned -1 could be interpreted as a huge positive index, bypassing the <0 guard and leading to out-of-bounds access when NVML devices are not
visible to CUDA (e.g., under CUDA_VISIBLE_DEVICES remapping)

mesutoezdil:
Hi,

We are going through open pull requests to keep things manageable. This PR was opened before 2026 and has not been merged or updated recently.

Please reply by 2026-07-13 to confirm this is still relevant and based on the current codebase. If we do not hear back, we will close this PR. You are welcome to reopen it or submit a new one when ready.

Thank you for your contribution to HAMi.

PR #133, opened 2025-10-11
fix wrong name

huazq:
src/cuda/hook.c:137 use wrong name "cucuMemImportFromShareableHandle"


archlitchi:
It has been merged in https://github.com/Project-HAMi/HAMi-core/pull/136

PR #160, opened 2026-03-17
Fix SIGUSR1/SIGUSR2 handler clobbering breaking JVM processes

charford:
## Problem

`libvgpu.so` registers signal handlers for `SIGUSR1` and `SIGUSR2` using `signal()`, which overwrites any previously installed handlers without saving them. This causes JVM processes to crash with `SIGSEGV` in `Monitor::wait()` because the JVM uses `SIGUSR1`/`SIGUSR2` internally for GC safepoints and thread management.

Observed on HAMi volcano-vgpu nodes (hami-core mode) when running PyTorch jobs with a JVM component — the crash occurs at startup before CUDA initializes.

Additional risks from the current implementation:
- `libvgpu.so` intercepts `dlsym()`, which the JVM also uses for native library loading
- The `ENSURE_RUNNING()` spin loop can cause a deadlock if a Java thread holding a JVM monitor gets suspended

## Fix

Replace `signal()` calls in `init_proc_slot_withlock()` with `sigaction()`, saving the previous handlers into static `struct sigaction` variables and chaining to them in `sig_swap_stub` / `sig_restore_stub`. This ensures other runtimes (JVM, Python, etc.) can still process `SIGUSR1`/`SIGUSR2` correctly.

## Changes

- `src/multiprocess/multiprocess_memory_limit.c`: Replace `signal()` with `sigaction()`, chain to previous handlers

## Testing

- Verified JVM workloads no longer crash with `SIGSEGV` in `Monitor::wait()` when running under `libvgpu.so` injection on HAMi volcano-vgpu nodes (prod-ltx1-k8s-1, prod-ltx1-k8s-2)

Closes #161

### DIFF
--- a/src/multiprocess/multiprocess_memory_limit.c
+++ b/src/multiprocess/multiprocess_memory_limit.c
@@ -11,6 +11,7 @@
 #include <unistd.h>
 #include <time.h>
 #include <signal.h>
+#include <string.h>
 
 #include <assert.h>
 #include <cuda.h>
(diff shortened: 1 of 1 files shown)


mesutoezdil:

Issue #167, opened 2026-03-26
[Bug] nvmlInit permanent failure in Burstable QoS pods due to pthread_once poisoning


maishivamhoo123:
Description
In Kubernetes 1.27+, Burstable Pods (where limits do not equal requests) can enter a resize: InProgress state during startup. When this happens, Kubernetes temporarily locks the GPU device files (/dev/nvidia*).
Related Issue:-  https://github.com/Project-HAMi/HAMi/issues/1704

The Problem:
HAMi-core uses pthread_once to initialize GPU maps. If the first attempt to start the GPU fails (because the device is locked), HAMi-core still marks the initialization as "Done." When the pod tries again a second later, HAMi-core thinks it is already finished and skips the setup. This causes the GPU to stay broken ("Unknown Error") for the entire life of the pod.

Root Cause Analysis
First Call: nvmlInit is called while the device is locked.

Driver Failure: The real NVIDIA library returns NVML_ERROR_UNKNOWN.

State Poisoning: Even though it failed, pthread_once marks the internal HAMi setup as "Finished."

Permanent Failure: Every future call to the GPU fails because the internal HAMi maps were never actually created.

Proposed Fix
We must only mark the initialization as "Done" if the NVIDIA driver actually returns NVML_SUCCESS. We should replace pthread_once with a manual check and a Mutex lock so it can be retried.

Suggested Code Fix
// 1. Add a manual flag and a lock at the top of the file

```
static pthread_mutex_t g_init_mutex = PTHREAD_MUTEX_INITIALIZER;
static int g_post_init_complete = 0;

```
// 2. Update the nvmlInit function
```
nvmlReturn_t nvmlInit(void) {
    // Keep preInit as is (it just sets up library paths)
    pthread_once(&init_virtual_map_pre_flag, (void(*)(void))nvml_preInit);

    // Call the real NVIDIA driver
    nvmlReturn_t res = NVML_OVERRIDE_CALL(nvml_library_entry, nvmlInit_v2);

    // FIX: Only run postInit if the driver actually succeeded!
    if (res == NVML_SUCCESS) {
        pthread_mutex_lock(&g_init_mutex);
        if (!g_post_init_complete) {
            nvml_postInit();
            g_post_init_complete = 1;
        }
        pthread_mutex_unlock(&g_init_mutex);
    }
    
    return res;
}
```
(Note: The same logic should be applied to nvmlInit_v2 and nvmlInitWithFlags in the same file.)

Environment Information
[...]

PR #168, opened 2026-03-26
fix(nvml): allow retrying nvmlInit to prevent permanent failure state

maishivamhoo123:
## Description
Fixes a critical edge case where GPU initialization permanently fails in Kubernetes 1.27+ Burstable pods (e.g., when device files are temporarily locked during `resize: InProgress`).
Issue Related : #167 and https://github.com/Project-HAMi/HAMi/issues/1704#issuecomment-4115925203

**The Bug:**
Because `nvml_postInit` used `pthread_once`, it was marked as "completed" even if the initial `nvmlInit` driver call failed. If an application retried the initialization later, `nvml_postInit` was bypassed, leaving the virtual GPU map permanently broken.

**The Fix:**
* Replaced `pthread_once` with a `pthread_mutex_t` and a `volatile int` flag.
* Updated `nvmlInit`, `nvmlInit_v2`, and `nvmlInitWithFlags` to only run and mark `nvml_postInit()` as complete **if** the NVIDIA driver returns `NVML_SUCCESS`.
* Applications can now safely retry initialization without poisoning the state.

## Changes Made
- **`src/nvml/hook.c`**: Replaced `pthread_once` with mutex-guarded success checks for post-initialization.
- **`test/test_nvml_init.c`**: Added a unit test to verify the retry logic.

## Checklist
- [x] Code follows the project's coding guidelines.
- [x] Tested locally using `make build-in-docker` (passes cleanly).
- [x] Added unit tests to verify the retry behavior.


maishivamhoo123:
@DSFans2014  , @archlitchi @Shouren and @team can you please review these PR?

archlitchi:
i got the issue, but i wonder why replace pthread_once, we can simply check NVML_SUCCESS before calling that

maishivamhoo123:
> i got the issue, but i wonder why replace pthread_once, we can simply check NVML_SUCCESS before calling that

Thank you @archlitchi for your review , yes you are right we should simply check NVML_SUCCESS before calling that ,  When I initially debugged this, I saw that pthread_once was rigidly locking our state on failure. Because pthread_once doesn't have a reset mechanism, my immediate thought was that the tool itself was too inflexible for a retry loop. Therefore, I focused entirely on writing a manual mutex and boolean flag to give us explicit control over the state. What I completely missed was that we didn't need to replace pthread_once, we just needed to reposition it. I will fix it.

Just to confirm, I will apply this exact same NVML_SUCCESS wrapper to nvmlInit, nvmlInit_v2, and nvmlInitWithFlags. Does that sound good to you?

archlitchi (approved):
/lgtm

fishman:
I've been wondering would this be a bit faster with an atomic CAS solution? not sure if that kind of performance matters here.

EDIT: never mind, I noticed you already removed the mutex requirement.

archlitchi:
Hi, looks like we got an issue during build, could you fix that?

mesutoezdil:

mesutoezdil:
resolve conflicts pls

maverick-woo:
This has already been fixed by #223. Please close this PR. Thanks!

Issue #169, opened 2026-04-13
support vllm 0.18.0


peachest:
When running vllm 0.18.0 with HAMi 2.8.0, an error occurs before the model is loaded. The error message is as follows:
```shell
[HAMI-core ERROR (pid:565 thread=140267266266944 multiprocess_memory_limit.c:831)]: Illegal device id: 850725928
!!!!!!! Segfault encountered !!!!!!!
  File "/libvgpu/src/multiprocess/multiprocess_memory_limit.c", line 833, in get_current_device_memory_limit
  File "/libvgpu/src/allocator/allocator.c", line 44, in oom_check
  File "/libvgpu/src/cuda/memory.c", line 582, in cuMemCreate
  File "misc/cudawrap.cc", line 92, in ncclCuMemHostEnable()
  File "misc/cudawrap.cc", line 202, in ncclCuMemHostEnable()
  File "misc/cudawrap.cc", line 280, in initOnceFunc
  File "misc/cudawrap.cc", line 298, in ncclCudaLibraryInit()
  File "/dvs/p4/build/sw/gpgpu/nccl/gitfusion/stable/src/init.cc", line 1852, in ncclCommInitRank
  File "<unknown>", line 0, in _PyObject_MakeTpCall
  File "<unknown>", line 0, in _PyEval_EvalFrameDefault
  File "<unknown>", line 0, in _PyObject_FastCallDictTstate
  File "<unknown>", line 0, in _PyObject_MakeTpCall
  File "<unknown>", line 0, in _PyEval_EvalFrameDefault
  File "<unknown>", line 0, in _PyObject_FastCallDictTstate
  File "<unknown>", line 0, in _PyObject_MakeTpCall
  File "<unknown>", line 0, in _PyEval_EvalFrameDefault
  File "<unknown>", line 0, in _PyObject_FastCallDictTstate
  File "<unknown>", line 0, in _PyObject_MakeTpCall
  File "<unknown>", line 0, in _PyEval_EvalFrameDefault
  File "<unknown>", line 0, in _PyObject_FastCallDictTstate
  File "<unknown>", line 0, in _PyObject_Call
  File "<unknown>", line 0, in _PyEval_EvalFrameDefault
  File "<unknown>", line 0, in PyEval_EvalCode
  File "<unknown>", line 0, in PyRun_StringFlags
  File "<unknown>", line 0, in PyRun_SimpleStringFlags
  File "<unknown>", line 0, in Py_RunMain
[...]

PR #172, opened 2026-04-16
fix: segfault in cuMemCreate hook when cuCtxGetDevice fails

astrox131:
## Problem

We ran into a segfault when loading models with `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`. This setting makes PyTorch use `cuMemCreate`/`cuMemMap` (VMM APIs) instead of `cuMemAlloc`.

In our case, `transformers` loads weights using a `ThreadPoolExecutor`. On worker threads where the CUDA context hasn't been established yet, `cuCtxGetDevice` can return `CUDA_ERROR_INVALID_CONTEXT` (201). The current `cuMemCreate` hook calls `cuCtxGetDevice` via `CUDA_OVERRIDE_CALL` which discards the return value, so the `dev` variable stays uninitialized. This garbage value gets passed into `oom_check` → `get_current_device_memory_limit`, leading to an out-of-bounds access and segfault.

We confirmed `cuCtxGetDevice` returns 201 on worker threads:

```
=== Main thread (after torch.cuda.init) ===
  cuCtxGetDevice ret=0, dev=0       ← OK

=== Worker threads (before any CUDA op) ===
  thread 0: cuCtxGetDevice ret=201, dev=-999  ← CUDA_ERROR_INVALID_CONTEXT
  thread 1: cuCtxGetDevice ret=201, dev=-999
  thread 2: cuCtxGetDevice ret=201, dev=-999
  thread 3: cuCtxGetDevice ret=201, dev=-999
```

This appears to be the same issue reported in #169.

## Proposed fix

```diff
     CUdevice dev;
-    CUDA_OVERRIDE_CALL(cuda_library_entry, cuCtxGetDevice, &dev);
-    if (oom_check(dev, size)) {
+    int do_oom_check = (prop->location.type == CU_MEM_LOCATION_TYPE_DEVICE);
+    if (do_oom_check && cuCtxGetDevice(&dev) != CUDA_SUCCESS) {
+        dev = prop->location.id;
+    }
+    if (do_oom_check && oom_check(dev, size)) {
         return CUDA_ERROR_OUT_OF_MEMORY;
     }
```

- Check `cuCtxGetDevice` return value. When it fails, fall back to `prop->location.id` (the target device from the allocation request) so the OOM check still works correctly.
- Only run OOM check for `CU_MEM_LOCATION_TYPE_DEVICE` allocations, since GPU memory limits don't apply to other location types.

Happy to adjust the approach if maintainers have a different preference.

## Test

Added `test/test_mem_create.c` — spawns 4 threads that call `cuMemCreate`. Without the fix this segfaults; with the fix all threads complete normally.

## Verification

Tested on H100 80GB with HAMI v2.7.1 (HAMi-core ec8979d):

| Test | Before fix | After fix |
|------|-----------|-----------|
| Python + `expandable_segments:True` | Segfault (exit 139) | OK (exit 0) |
| Python without `expandable_segments` | OK | OK |
| `test_mem_create` (C, 4 threads, no context) | Segfault | OK, all threads pass |
| `test_alloc` | OK | OK |
| `test_alloc_managed` | OK | OK |
| `test_alloc_pitch` | OK | OK |

## Reproducer (Python, optional)

<details>
<summary>Click to expand</summary>

```python
# pip install torch transformers safetensors
# Crashes:  PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True python repro.py
# Works:    python repro.py
import argparse, json, os, tempfile, torch
from safetensors.torch import save_file

BERT_CONFIG = {
    "architectures": ["BertModel"], "model_type": "bert",
    "hidden_size": 4096, "intermediate_size": 16384,
    "num_hidden_layers": 20, "num_attention_heads": 32,
    "max_position_embeddings": 512, "vocab_size": 30522,
    "type_vocab_size": 2, "hidden_act": "gelu",
    "layer_norm_eps": 1e-12, "hidden_dropout_prob": 0.1,
    "attention_probs_dropout_prob": 0.1,
}

def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("--single-thread", action="store_true")
    args = parser.parse_args()
    print(f"PyTorch {torch.__version__}, CUDA {torch.version.cuda}")
    print(f"PYTORCH_CUDA_ALLOC_CONF={os.environ.get('PYTORCH_CUDA_ALLOC_CONF', '(unset)')}")
    if args.single_thread:
        import transformers.core_model_loading as cml
        cml.GLOBAL_WORKERS = 0
    with tempfile.TemporaryDirectory() as d:
        with open(os.path.join(d, "config.json"), "w") as f:
            json.dump(BERT_CONFIG, f)
        from transformers import AutoModel, AutoConfig
        model = AutoModel.from_config(AutoConfig.for_model(**BERT_CONFIG))
        save_file(model.state_dict(), os.path.join(d, "model.safetensors"))
        del model
        AutoModel.from_pretrained(d, device_map="cuda")
        print("OK")

if __name__ == "__main__":
    main()
```

</details>

## Environment

- HAMI v2.7.1 (HAMi-core ec8979d)
- NVIDIA H100 80GB HBM3, Driver 570.x, CUDA 12.8
- PyTorch 2.9.1+cu128, transformers 4.52.4

### DIFF
--- a/src/cuda/memory.c
+++ b/src/cuda/memory.c
@@ -160,7 +160,7 @@ CUresult cuMemAllocManaged(CUdeviceptr* dptr, size_t bytesize, unsigned int flag
     LOG_DEBUG("cuMemAllocManaged dptr=%p bytesize=%ld",dptr,bytesize);
     ENSURE_RUNNING();
     CUdevice dev;
-    CUDA_OVERRIDE_CALL(cuda_library_entry,cuCtxGetDevice,&dev);
+    CHECK_DRV_API(cuCtxGetDevice(&dev));
     if (oom_check(dev,bytesize)){
         return CUDA_ERROR_OUT_OF_MEMORY;
     }
@@ -178,7 +178,7 @@ CUresult cuMemAllocPitch_v2(CUdeviceptr* dptr, size_t* pPitch, size_t WidthInByt
     size_t bytesize = guess_pitch * Height;
     ENSURE_RUNNING();
     CUdevice dev;
-    CUDA_OVERRIDE_CALL(cuda_library_entry,cuCtxGetDevice,&dev);
+    CHECK_DRV_API(cuCtxGetDevice(&dev));
     if (oom_check(dev,bytesize)){
         return CUDA_ERROR_OUT_OF_MEMORY;
     }
@@ -588,8 +588,11 @@ CUresult cuMemCreate ( CUmemGenericAllocationHandle* handle, size_t size, const 
     LOG_INFO("cuMemCreate:%lld:%d", size, prop->location.id);
     ENSURE_RUNNING();
     CUdevice dev;
-    CUDA_OVERRIDE_CALL(cuda_library_entry, cuCtxGetDevice, &dev);
-    if (oom_check(dev, size)) {
+    int do_oom_check = (prop->location.type == CU_MEM_LOCATION_TYPE_DEVICE);
+    if (do_oom_check && cuCtxGetDevice(&dev) != CUDA_SUCCESS) {
+        dev = prop->location.id;
+    }
+    if (do_oom_check && oom_check(dev, size)) {
         return CUDA_ERROR_OUT_OF_MEMORY;
     }
     CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,
--- a/test/test_mem_create.c
+++ b/test/test_mem_create.c
@@ -1,0 +1,77 @@
+#include <stdio.h>
+#include <cuda.h>
+#include <pthread.h>
+
+#include "test_utils.h"
+
+#define NUM_THREADS 4
+#define ALLOC_SIZE  (64 * 1024 * 1024)
+
+static CUdevice g_device;
+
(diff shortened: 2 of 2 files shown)


archlitchi:
Yes, that cause vllm error when using multiple GPUs, we have a lot of users reporting this issue, thanks for fixing this

archlitchi:
could you add the CUDA_SUCCESS check to other similar cuMemAllocate methods(e.g. cuMemAllocManaged) as well?

astrox131:
Hi maintainers, this PR is ready for review when you have a chance.

A quick note on the `add_chunk_only` / `remove_chunk_only` path: these are only called after the real CUDA driver's `cuMemCreate` succeeds (`if (res == CUDA_SUCCESS)`), which means the driver has already established the CUDA context on the calling thread by that point. So `cuCtxGetDevice` inside `add_chunk_only` should work correctly in this scenario.

Happy to adjust anything based on your feedback. Thanks!

archlitchi (approved):
/lgtm

PR #173, opened 2026-04-17
feat: make build.sh compatible with dash

Shouren:
Refactored `build.sh` to replace bash-specific syntax with POSIX-compliant equivalents, making the script portable across shells such as `sh` and `dash`.


archlitchi (approved):
/lgtm

PR #176, opened 2026-04-21
fix(test): test_mem_create failing due to uninitialized NVML shutdown

anandj91:
test_mem_create was calling nvmlShutdown() at the end of main() without ever initializing NVML.
Also removing unnecessary sleep in test_runtime_launch


archlitchi (approved):
/lgtm

PR #184, opened 2026-04-27
fix: make build.sh resilient to missing git

spencercjh:
## Summary
- `build.sh` used `set -e` with `$(git describe ...)` and a `$?` check that was never reached when git is unavailable
- This was discovered during a HAMi Docker build, where the `nvbuild` stage only installs `cmake` (not `git`), causing `build.sh` to fail immediately with "git: command not found"
- Replaced with a single `|| echo "unknown"` fallback that works correctly under `set -e`
- Also added `.idea`, `.vscode`, `.claude` to `.gitignore`

## Test plan
- [x] Script still works normally when git is available
- [x] Script falls back to `"unknown"` when git is missing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 build
 cmake-build-debug
+.idea
+.vscode
+.claude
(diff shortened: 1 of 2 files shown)


archlitchi (approved):
/lgtm

PR #185, opened 2026-04-28
fix: handle empty GIT_BRANCH in CMakeLists.txt when HEAD is detached

spencercjh:
## Summary
- When git is installed but HEAD is detached (e.g., submodule checkout in CI), `git symbolic-ref --short -q HEAD` returns nothing, leaving `GIT_BRANCH` empty.
- CMake's `string(REPLACE)` on an empty variable causes: `string sub-command REPLACE requires at least four arguments.`
- This prevents Makefile generation and causes the build to fail.

## Root cause
The dash-compatibility PR (#173) added quotes to `string(REPLACE)` arguments, but did not guard against empty `GIT_BRANCH`. When git is NOT installed (`GIT_FOUND=false`), the block is skipped entirely — so this bug was hidden until git was installed in the build environment.

## Fix
Skip `string(REPLACE)` when `GIT_BRANCH` is empty, and default to `"unknown"`.

Related: https://github.com/Project-HAMi/HAMi/pull/1782 (CI failure caused by this bug)


spencercjh:
cc @Shouren

spencercjh:
Associated PR: https://github.com/Project-HAMi/HAMi/pull/1782

archlitchi (approved):
/lgtm

PR #186, opened 2026-04-28
fix cuMemcreate cause core dump and get memory usage 0 after allocate twice

maverick123123:
**Summary**                                                                                                                                                                                                            
                                                                                                                                                                                                                     
  Fix two bugs introduced by prior refactoring: cuMemCreate causes a core dump due to improper CUDA context handling, and memory usage reporting always shows 0 when allocating more than once.                      
                                                                                                                                                                                                                     
  Changes                                                                                                                                                                                                            
                  
  1. **Fix core dump in cuMemCreate and other allocation paths**                                                                                                                                                         
                  
  add_chunk_only and remove_chunk_only were internally calling cuCtxGetDevice to determine the current device, but this call can fail or return incorrect results when the CUDA context is in certain states (e.g.,  
  after cuMemCreate, which may not have a bound context). This caused a core dump on cuMemCreate.
                                                                                                                                                                                                                     
  Fix: Pass CUdevice dev as a parameter from the caller instead of querying it internally.                                                                                                                           
  
  - add_chunk_only now takes a CUdevice dev parameter — all callers (cuMemAllocManaged, cuMemAllocPitch_v2, cuMemCreate) pass the device they already know                                                           
  - remove_chunk_only now reads the device from the stored allocated_device_memory_struct.dev field, which is populated at allocation time by the INIT_ALLOCATED_LIST_ENTRY macro
                                                                                                                                                                                                                     
  2. **Fix memory usage always showing 0 after the second allocation**                                                                                                                                                   
                                                                                                                                                                                                                     
  INIT_ALLOCATED_LIST_ENTRY previously did not store the CUdevice in the entry. When remove_chunk_only ran (e.g., on the second free), it called cuCtxGetDevice which could return a stale or incorrect device,      
  causing rm_gpu_device_memory_usage to decrement the wrong counter — leaving the real counter non-zero and causing subsequent get_memory_usage calls to show 0.
                                                                                                                                                                                                                     
  Fix: Added CUdevice dev field to allocated_device_memory_struct and set it in INIT_ALLOCATED_LIST_ENTRY. remove_chunk_only uses val->entry->dev instead of querying the context.                                   
  
  3. Atomic safety in clear_proc_slot_nolock                                                                                                                                                                         
                  
  Direct struct assignment on shared region process slots with _Atomic members is non-atomic and can cause torn reads.                                                                                               
                  
  Fix: Added copy_proc_slot_atomic — copies each member individually using atomic loads/stores. When compacting the slot array and the moved slot is my_slot, the pointer is updated and the vacated slot is zeroed  
  out atomically to prevent use-after-free by concurrent readers.

### DIFF
--- a/src/allocator/allocator.c
+++ b/src/allocator/allocator.c
@@ -108,7 +108,7 @@ int add_chunk(CUdeviceptr *address, size_t size) {
         return CUDA_ERROR_OUT_OF_MEMORY;
     
     allocated_list_entry *e;
-    INIT_ALLOCATED_LIST_ENTRY(e,addr,size);
+    INIT_ALLOCATED_LIST_ENTRY(e, addr, size, dev);
     if (size <= IPCSIZE)
         res = CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemAlloc_v2,&e->entry->address,size);
     else{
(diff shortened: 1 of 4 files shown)


archlitchi (approved):
/lgtm

PR #188, opened 2026-04-28
fix(cuda,allocator): initialize dev in 6 hooks reading uninitialised stack on cuCtxGetDevice failure

### LINKED ISSUE #187: Bug: cuMemCreate hook reads uninitialized dev when   prop->location.type != CU_MEM_LOCATION_TYPE_DEVICE
[repro.c](https://github.com/user-attachments/files/27178572/repro.c)

# Bug: `cuMemCreate` hook reads uninitialized `dev` when `prop->location.type != CU_MEM_LOCATION_TYPE_DEVICE`

## Summary

In `src/cuda/memory.c`, the `cuMemCreate` hook calls `add_chunk_only(*handle, size, dev)` with `dev` left uninitialized whenever `prop->location.type` is not `CU_MEM_LOCATION_TYPE_DEVICE`. The garbage value flows down into `set_current_device_memory_limit`, which logs `Illegal device id: <random>` and writes out-of-bounds into `region_info.shared_region->limit[dev]`.

This blocks any container that uses CUDA Driver API VMM (`cuMemCreate` + `cuMemAddressReserve` + `cuMemMap`) with a non-DEVICE allocation location — which is the path `ggml-cuda` takes when staging a virtual address pool before binding it to a specific device. So basically `llama.cpp`, every `llama-server` build with VMM on, and the Lucebox / DFlash speculative-decoding stack all crash on a HAMi-managed pod.

## Steps to reproduce

Hardware: NVIDIA RTX 5090 Laptop GPU (sm_120 Blackwell consumer). The bug isn't hardware-specific — same pattern should hit on any GPU under HAMi as long as the workload uses `cuMemCreate` with a non-DEVICE location.

Stack: HAMi 2.5.1 (deployed as part of Olares 1.12.5), `nvidia/cuda:13.0.0-devel-ubuntu22.04`, `ggml-cuda` built with default `GGML_CUDA=ON` (i.e. with VMM enabled).

Pod logs at startup:

```
ggml_cuda_init: found 1 CUDA devices (Total VRAM: 24463 MiB):
  Device 0: NVIDIA GeForce RTX 5090 Laptop GPU, compute capability 12.0,
            VMM: yes, VRAM: 24463 MiB
[HAMI-core ERROR (pid:1 thread=124663410515968 multiprocess_memory_limit.c:846)]:
Illegal device id: -644371744
```

The `-644371744` value changes between runs — whatever was on the stack at the previous frame. I confirmed it's not a build, env or driver issue: same pattern across image rebuilds, `LD_LIBRARY_PATH` permutations, `CUDA_DEVICE_MEMORY_LIMIT_0` settings, and on two different driver versions.

Minimal C reproducer:

```c
#include <stdio.h>
#include <cuda.h>

int main(void) {
    cuInit(0);
    CUdevice device; cuDeviceGet(&device, 0);
    CUcontext ctx;   cuCtxCreate(&ctx, 0, device);

    CUmemAllocationProp prop = {0};
    prop.type             = CU_MEM_ALLOCATION_TYPE_PINNED;
    prop.location.type    = CU_MEM_LOCATION_TYPE_HOST_NUMA;  // not DEVICE
    prop.location.id      = 0;

    size_t granularity;
    cuMemGetAllocationGranularity(&granularity, &prop,
                                  CU_MEM_ALLOC_GRANULARITY_MINIMUM);
    size_t size = ((1<<20) + granularity - 1) / granularity * granularity;

    CUmemGenericAllocationHandle handle;
    CUresult res = cuMemCreate(&handle, size, &prop, 0);
    printf("cuMemCreate returned %d\n", res);
    cuCtxDestroy(ctx);
    return 0;
}
```

`nvcc -lcuda repro.c -o repro` and run it inside a HAMi-managed pod — `[HAMI-core ERROR ...]: Illegal device id: <random>` shows up immediately.

## Root cause

`src/cuda/memory.c`, `cuMemCreate` hook around line 1009:

```c
CUresult cuMemCreate(CUmemGenericAllocationHandle* handle, size_t size,
                     const CUmemAllocationProp* prop, unsigned long long flags) {
    LOG_INFO("cuMemCreate:%lld:%d", size, prop->location.id);
    ENSURE_RUNNING();
    CUdevice dev;                                          // (a) not initialised
    int do_oom_check = (prop->location.type == CU_MEM_LOCATION_TYPE_DEVICE);
    if (do_oom_check && cuCtxGetDevice(&dev) != CUDA_SUCCESS) {
        dev = prop->location.id;
    }
    if (do_oom_check && oom_check(dev, size)) {
        return CUDA_ERROR_OUT_OF_MEMORY;
    }
    CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,
        cuMemCreate, handle, size, prop, flags);
    if (res == CUDA_SUCCESS) {
        add_chunk_only(*handle, size, dev);                // (b) uses (a) uninitialised
    }
    return res;
}
```

When `prop->location.type != CU_MEM_LOCATION_TYPE_DEVICE`, the entire `if (do_oom_check && ...)` block is skipped, so `dev` stays whatever was on the stack. The unconditional `add_chunk_only(*handle, size, dev)` at the bottom forwards that garbage to `set_current_device_memory_limit`:

```c
// src/multiprocess/multiprocess_memory_limit.c
int set_current_device_memory_limit(const int dev, size_t newlimit) {
    ensure_initialized();
    if (dev < 0 || dev >= CUDA_DEVICE_MAX_COUNT) {
        LOG_ERROR("Illegal device id: %d", dev);
    }
    LOG_INFO("dev %d new limit set to %ld",dev,newlimit);
    region_info.shared_region->limit[dev]=newlimit;        // OOB write
    return 0;
}
```

Two issues actually:

1. `cuMemCreate` reads uninitialised `dev` (the visible symptom).
2. `set_current_device_memory_limit` logs the error and then *proceeds* to `region_info.shared_region->limit[dev] = newlimit` — that's an out-of-bounds write into shared memory, latent corruption for any other process attached to the same `region_info`.

The real-world path that triggers (1): `ggml-cuda` reserves a virtual address pool via `cuMemCreate` with `CU_MEM_LOCATION_TYPE_HOST_NUMA` for staging before binding to a specific device. Standard CUDA Driver API VMM usage, see [NVIDIA's intro post](https://developer.nvidia.com/blog/introducing-low-level-gpu-virtual-memory-management/).

## Suggested fix

Two-line minimal fix at the `cuMemCreate` site — don't track non-DEVICE allocations against per-device memory limits since they don't consume device VRAM:

```c
CUdevice dev = prop->location.id;                          // initialised up-front
int do_oom_check = (prop->location.type == CU_MEM_LOCATION_TYPE_DEVICE);
if (do_oom_check && cuCtxGetDevice(&dev) != CUDA_SUCCESS) {
    dev = prop->location.id;
}
if (do_oom_check && oom_check(dev, size)) {
    return CUDA_ERROR_OUT_OF_MEMORY;
}
CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,
    cuMemCreate, handle, size, prop, flags);
if (res == CUDA_SUCCESS && do_oom_check) {                 // ← skip non-DEVICE
    add_chunk_only(*handle, size, dev);
}
return res;
```

While we're at it, `set_current_device_memory_limit` should bail out instead of writing out of bounds:

```c
int set_current_device_memory_limit(const int dev, size_t newlimit) {
    ensure_initialized();
    if (dev < 0 || dev >= CUDA_DEVICE_MAX_COUNT) {
        LOG_ERROR("Illegal device id: %d", dev);
        return -1;                                         // ← was missing
    }
    LOG_INFO("dev %d new limit set to %ld", dev, newlimit);
    region_info.shared_region->limit[dev] = newlimit;
    return 0;
}
```

I'm happy to send a PR if that's the direction you'd take, or to test a different approach if you'd rather track non-DEVICE allocs in a separate accounting path.

## Why this matters

VMM is on by default in `ggml-cuda` since [llama.cpp PR #11446](https://github.com/ggml-org/llama.cpp/pull/11446) (Jan 2026), which means every recent `llama-server` build hits this bug on a HAMi-managed pod. The current workaround is to rebuild with `-DGGML_CUDA_NO_VMM=ON`, which works but loses the VMM benefits (notably ~3.5× KV cache savings via TurboQuant TQ3_0 paths).

A fix in HAMi-core would unblock upstream images (`ggml-org/llama.cpp:server-cuda13-*`, `vllm/vllm-openai` recent builds, the Lucebox `lucebox-hub` consumer-tuned fork) without users having to rebuild anything.

## Environment

- HAMi-core: `master`, commit `ec8979d` (the submodule pin in HAMi `master`)
- HAMi: 2.5.1 (deployed via the `hami-2.5.1` Helm chart in `kube-system`)
- Olares: 1.12.5
- GPU: NVIDIA RTX 5090 Laptop GPU, compute capability 12.0 (sm_120 Blackwell consumer)
- Driver: 580.65.06, CUDA 13.0.48
- Runtime: containerd via k3s 1.31
- Reproducer images on Docker Hub:
  - `aamsellem/lucebox-qwen36-blackwell:1.0.0` — VMM enabled, triggers the bug
  - `aamsellem/lucebox-qwen36-blackwell:1.1.0` — same code, `-DGGML_CUDA_NO_VMM=ON`, no bug

Happy to send extra logs, the full reproducer pod manifest, or a PR.

aamsellem:
Fixes #187.

### What

`cuCtxGetDevice(&dev)` is called in 6 different hooks across `src/cuda/memory.c` and `src/allocator/allocator.c` without checking the return value. Whenever `cuCtxGetDevice` fails (e.g. a CUDA call comes from a thread that hasn't pushed a context yet, which is common in async / multi-thread engines), `dev` is left uninitialised on the stack and forwarded to:

- `oom_check(dev, size)` → reads garbage as device id, indexes into shared region
- `add_chunk_only(*handle, size, dev)` → stores garbage as the entry's device
- `add_gpu_device_memory_usage(getpid(), dev, allocsize, 2)` → indexes into shared region
- `rm_gpu_device_memory_usage(getpid(), dev, t_size, 2)` → same
- `set_current_device_memory_limit(dev, newlimit)` → logs `Illegal device id: <random>` and writes OOB into `region_info.shared_region->limit[dev]`

The visible symptom in pod logs is:

```
[HAMI-core ERROR ...]: Illegal device id: -644371744
```

with the random integer changing between runs. The silent symptom is shared-region corruption that can affect other processes attached to the same `region_info`.

This blocks any container that does multi-threaded / early CUDA allocation under HAMi — notably `ggml-cuda` (used by `llama.cpp` `llama-server`, the Lucebox / DFlash speculative-decoding stack), and any vLLM build with async scheduling.

### How

Two commits.

**1. `cuMemCreate` (`src/cuda/memory.c`)** — initialise `dev` from `prop->location.id` up-front, and only call `add_chunk_only` when the allocation is actually DEVICE-pinned (`do_oom_check == true`). Non-DEVICE allocations don't consume per-device VRAM so they shouldn't count against per-device limits.

Also tightens `set_current_device_memory_limit`: it now returns `-1` early instead of falling through to an OOB write when `dev` is out of range.

**2. `src/allocator/allocator.c`** — same pattern fix in 5 more sites:

| Site | Behaviour |
|---|---|
| `oom_check` (dev == -1 branch) | bail out gracefully, return 0 (skip OOM accounting) |
| `add_chunk` | bail out, return CUDA_SUCCESS (alloc still tracked by the lib) |
| `remove_chunk` | skip per-device usage tracking, leave the lib free intact |
| `remove_chunk_async` | same |
| `add_chunk_async` | bail out, return CUDA_SUCCESS |

Strategy is the same in all five: initialise `dev = -1`, check the return code of `cuCtxGetDevice`, and on failure skip per-device tracking gracefully. The underlying CUDA allocation/free still happens; we just don't bill it against a device limit if we don't know which device it belongs to. Safer than making up a device id and writing OOB into the shared region.

Also drops two redundant `cuCtxGetDevice()` calls in `add_chunk` and `add_chunk_async` — `dev` is already set at the top of each function, re-querying after the allocation just multiplies the chance of the same race.

### Test

Reproducer in #187. Quick manual check:

```c
// repro.c — should print 0 (CUDA_SUCCESS) and not log "Illegal device id"
#include <stdio.h>
#include <cuda.h>
int main(void) {
    cuInit(0);
    CUdevice device; cuDeviceGet(&device, 0);
    CUcontext ctx;   cuCtxCreate(&ctx, 0, device);
    CUmemAllocationProp prop = {0};
    prop.type             = CU_MEM_ALLOCATION_TYPE_PINNED;
    prop.location.type    = CU_MEM_LOCATION_TYPE_HOST_NUMA;
    prop.location.id      = 0;
    size_t granularity;
    cuMemGetAllocationGranularity(&granularity, &prop, CU_MEM_ALLOC_GRANULARITY_MINIMUM);
    size_t size = ((1<<20) + granularity - 1) / granularity * granularity;
    CUmemGenericAllocationHandle handle;
    CUresult res = cuMemCreate(&handle, size, &prop, 0);
    printf("cuMemCreate returned %d\n", res);
    cuCtxDestroy(ctx);
    return 0;
}
```

Build: `nvcc -lcuda repro.c -o repro`. Run inside a HAMi-managed pod.

- Before: `[HAMI-core ERROR ...]: Illegal device id: <random>` then `cuMemCreate returned 0`.
- After: just `cuMemCreate returned 0`. No HAMi error, no OOB write.

End-to-end check: `aamsellem/lucebox-qwen36-blackwell:1.0.0` pod boots through `ggml_cuda_init` and reaches `llama-server is listening` on a sm_120 GPU, instead of crash-looping at the first VMM allocation.

### Notes

- Behavioural change for `cuMemCreate`: non-DEVICE allocations are no longer added to the chunk tracker. They never consumed device VRAM in the first place, so this lines up with what HAMi actually bills against. If maintainers prefer to track them in a separate accounting path instead, happy to rework — just say the word.
- Behavioural change for the allocator hooks: when `cuCtxGetDevice` fails, the per-device accounting is skipped, but the underlying CUDA call still runs. The shared region stays consistent rather than being written OOB. If maintainers prefer a different default (e.g. fall back to device 0), happy to switch.
- The bounds-check fix in `set_current_device_memory_limit` is technically a separate concern. I kept it in the same PR because it's a 1-line addition and the symptom (`Illegal device id` log followed by silent shared-region corruption) was confusing me during the initial debug. Easy to split if preferred.
- The same OOB-after-LOG_ERROR pattern still exists in the symmetric getters/setters (`get_current_device_sm_limit`, `set_current_device_sm_limit_scale`, `get_current_device_memory_limit`, `get_current_device_memory_monitor`, `get_current_device_memory_usage`). Those weren't on the crash path I was tracing so I left them alone — happy to do them in this PR or as a follow-up, just let me know.
- Tested on HAMi 2.5.1 / HAMi-core master at commit `b6afdd0`, driver 580.65.06, CUDA 13.0.48, RTX 5090 Laptop GPU (sm_120). Should be hardware-agnostic.

### Checklist

- [x] Tested on HAMi-managed pod with both VMM-enabled and VMM-disabled builds — VMM-enabled now boots, VMM-disabled unaffected.
- [x] Repro pre-fix and post-fix documented in #187.
- [ ] No new tests added — `src/multiprocess/` and `src/allocator/` don't currently ship with a test harness in this repo. Happy to add one if there's a pattern to follow.

### DIFF
--- a/src/allocator/allocator.c
+++ b/src/allocator/allocator.c
@@ -34,11 +34,15 @@ size_t round_up(size_t size, size_t unit) {
 }
 
 int oom_check(const int dev, size_t addon) {
-    CUdevice d;
-    if (dev==-1)
-        cuCtxGetDevice(&d);
-    else
+    CUdevice d = -1;
+    if (dev==-1) {
+        if (cuCtxGetDevice(&d) != CUDA_SUCCESS) {
+            LOG_WARN("oom_check: cuCtxGetDevice failed, skipping check");
+            return 0;
+        }
+    } else {
         d=dev;
+    }
     uint64_t limit = get_current_device_memory_limit(d);
     size_t _usage = get_gpu_memory_usage(d);
 
@@ -102,11 +106,14 @@ int add_chunk(CUdeviceptr *address, size_t size) {
     size_t addr=0;
     size_t allocsize;
     CUresult res = CUDA_SUCCESS;
-    CUdevice dev;
-    cuCtxGetDevice(&dev);
+    CUdevice dev = -1;
+    if (cuCtxGetDevice(&dev) != CUDA_SUCCESS) {
+        LOG_WARN("add_chunk: cuCtxGetDevice failed, skipping memory tracking");
+        return CUDA_SUCCESS;
+    }
     if (oom_check(dev,size))
         return CUDA_ERROR_OUT_OF_MEMORY;
-    
+
     allocated_list_entry *e;
     INIT_ALLOCATED_LIST_ENTRY(e, addr, size, dev);
     if (size <= IPCSIZE)
@@ -123,7 +130,6 @@ int add_chunk(CUdeviceptr *address, size_t size) {
     //uint64_t t_size;
     *address = e->entry->address;
     allocsize = size;
-    cuCtxGetDevice(&dev);
     add_gpu_device_memory_usage(getpid(), dev, allocsize, 2);
     return 0;
 }
@@ -168,9 +174,12 @@ int remove_chunk(allocated_list *a_list, CUdeviceptr dptr) {
             t_size=val->entry->length;
             cuMemoryFree(dptr);
             LIST_REMOVE(a_list,val);
-            CUdevice dev;
-            cuCtxGetDevice(&dev);
-            rm_gpu_device_memory_usage(getpid(), dev, t_size, 2);
+            CUdevice dev = -1;
+            if (cuCtxGetDevice(&dev) == CUDA_SUCCESS) {
+                rm_gpu_device_memory_usage(getpid(), dev, t_size, 2);
+            } else {
+                LOG_WARN("remove_chunk: cuCtxGetDevice failed, skipping memory tracking");
+            }
             return 0;
         }
     }
@@ -225,9 +234,12 @@ int remove_chunk_async(
             CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemFreeAsync,dptr,hStream);
             LIST_REMOVE(a_list,val);
             a_list->limit-=t_size;
-            CUdevice dev;
-            cuCtxGetDevice(&dev);
-            rm_gpu_device_memory_usage(getpid(),dev,t_size,2);
+            CUdevice dev = -1;
+            if (cuCtxGetDevice(&dev) == CUDA_SUCCESS) {
+                rm_gpu_device_memory_usage(getpid(),dev,t_size,2);
+            } else {
+                LOG_WARN("remove_chunk_async: cuCtxGetDevice failed, skipping memory tracking");
+            }
             return 0;
         }
     }
@@ -245,8 +257,11 @@ int add_chunk_async(CUdeviceptr *address, size_t size, CUstream hStream) {
     size_t addr=0;
     size_t allocsize;
     CUresult res = CUDA_SUCCESS;
-    CUdevice dev;
-    cuCtxGetDevice(&dev);
+    CUdevice dev = -1;
+    if (cuCtxGetDevice(&dev) != CUDA_SUCCESS) {
+        LOG_WARN("add_chunk_async: cuCtxGetDevice failed, skipping memory tracking");
+        return CUDA_SUCCESS;
+    }
     if (oom_check(dev,size))
         return -1;
 
@@ -273,13 +288,12 @@ int add_chunk_async(CUdeviceptr *address, size_t size, CUstream hStream) {
     if (poollimit != 0) {
         if (poollimit> device_allocasync->limit) {
             allocsize = (poollimit-device_allocasync->limit < size)? poollimit-device_allocasync->limit : size;
-            cuCtxGetDevice(&dev);
             add_gpu_device_memory_usage(getpid(), dev, allocsize, 2);
             device_allocasync->limit=device_allocasync->limit+allocsize;
             e->entry->length=allocsize;
         }else{
             e->entry->length=0;
-        } 
+        }
     }
     LIST_ADD(device_allocasync,e);
     return 0;
--- a/src/cuda/memory.c
+++ b/src/cuda/memory.c
@@ -587,7 +587,7 @@ CUresult cuMemAddressReserve(CUdeviceptr* ptr, size_t size,
 CUresult cuMemCreate ( CUmemGenericAllocationHandle* handle, size_t size, const CUmemAllocationProp* prop, unsigned long long flags ) {
     LOG_INFO("cuMemCreate:%lld:%d", size, prop->location.id);
     ENSURE_RUNNING();
-    CUdevice dev;
+    CUdevice dev = prop->location.id;
     int do_oom_check = (prop->location.type == CU_MEM_LOCATION_TYPE_DEVICE);
     if (do_oom_check && cuCtxGetDevice(&dev) != CUDA_SUCCESS) {
         dev = prop->location.id;
@@ -597,7 +597,7 @@ CUresult cuMemCreate ( CUmemGenericAllocationHandle* handle, size_t size, const 
     }
     CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,
         cuMemCreate, handle, size, prop, flags);
-    if (res == CUDA_SUCCESS) {
+    if (res == CUDA_SUCCESS && do_oom_check) {
         add_chunk_only(*handle, size, dev);
     }
     return res;
--- a/src/multiprocess/multiprocess_memory_limit.c
+++ b/src/multiprocess/multiprocess_memory_limit.c
@@ -1243,10 +1243,11 @@ int set_current_device_memory_limit(const int dev,size_t newlimit) {
     ensure_initialized();
     if (dev < 0 || dev >= CUDA_DEVICE_MAX_COUNT) {
         LOG_ERROR("Illegal device id: %d", dev);
+        return -1;
     }
     LOG_INFO("dev %d new limit set to %ld",dev,newlimit);
     region_info.shared_region->limit[dev]=newlimit;
-    return 0; 
+    return 0;
 }
 
 uint64_t get_current_device_memory_limit(const int dev) {


archlitchi:
Hi, this illegal device ID issue should have fixed in v2.8.1, have you tried that?

aamsellem:
Hi @archlitchi, thanks for the quick reply ! Just to make sure we're talking about the same thing — `v2.8.1` refers to the [`Project-HAMi/HAMi`](https://github.com/Project-HAMi/HAMi) wrapper release, right ? HAMi-core itself doesn't have releases, so I checked the master branch directly.

The bug is still present on `Project-HAMi/HAMi-core` master at the current HEAD (`e3c3477`). Lines 587-602 of `src/cuda/memory.c` :

```c
CUresult cuMemCreate ( CUmemGenericAllocationHandle* handle, size_t size, const CUmemAllocationProp* prop, unsigned long long flags ) {
    LOG_INFO("cuMemCreate:%lld:%d", size, prop->location.id);
    ENSURE_RUNNING();
    CUdevice dev;                                          // ← still uninitialized
    int do_oom_check = (prop->location.type == CU_MEM_LOCATION_TYPE_DEVICE);
    if (do_oom_check && cuCtxGetDevice(&dev) != CUDA_SUCCESS) {
        dev = prop->location.id;
    }
    // ...
    if (res == CUDA_SUCCESS) {
        add_chunk_only(*handle, size, dev);                // ← uninit dev when do_oom_check == false
    }
```

When `prop->location.type != CU_MEM_LOCATION_TYPE_DEVICE` (e.g. ggml-cuda using `CU_MEM_LOCATION_TYPE_HOST_NUMA` for its virtual address pool), `do_oom_check == false`, so the `if (...)` block that initialises `dev` is skipped entirely. Then `add_chunk_only(*handle, size, dev)` is called with whatever was on the stack.

Same pattern in 5 more sites in `src/allocator/allocator.c` (lines 39, 106, 172, 229, 249, 276) — all calling `cuCtxGetDevice(&dev)` with the return code ignored. If `cuCtxGetDevice` fails (which happens in async / multi-thread / early-init paths), `dev` ends up forwarded to `set_current_device_memory_limit` → `Illegal device id: <random>` + OOB write into `region_info.shared_region->limit[dev]`.

The minimal reproducer in #187 triggers the bug deterministically against current master. I built it against `e3c3477` to double-check before responding here.

Could you point me to the specific commit / PR that you think fixes this on HAMi (the wrapper) ? If there's a related fix on the wrapper side I'm happy to align my approach, but the uninitialised stack read in HAMi-core looks distinct.

mesutoezdil:

PR #192, opened 2026-05-09
fix/vllm eager mode cause core dump

maverick123123:
---                                                                                                                                                                                                                
  Summary
                                                                                                                                                                                                                     
  Fix two issues causing vLLM core dumps with HAMi-core: incorrect memory tracking in cuMemCreate, and NCCL CUDA graph capture segfault.
                                                                                                                                                                                                                     
  Changes                                                                                                                                                                                                            
                                                                                                                                                                                                                     
  1. Fix cuMemCreate memory tracking (src/cuda/memory.c)                                                                                                                                                             
                                                            
  cuMemCreate was calling add_chunk_only to track allocations even when OOM check was skipped (do_oom_check == false). This caused incorrect memory usage accounting because chunks were added to the tracking list  
  without proper validation.                                
                                                                                                                                                                                                                     
  - if (res == CUDA_SUCCESS) {                                                                                                                                                                                       
  + if (do_oom_check && res == CUDA_SUCCESS) {                                                                                                                                                                       
                                                                                                                                                                                                                     
  2. Fix NCCL CUDA graph capture segfault                                                                                                                                                                            
                                                                                                                                                                                                                     
  Symptom: vLLM with tensor-parallel-size=8 on 8x RTX 3090 segfaults in cuGraphAddDependencies_v2 during NCCL CUDA graph capture. Works fine with --enforce-eager.                                                   
                                                            
  Two root causes:                                                                                                                                                                                                   
                                                            
  2a. Allocator lock contention (src/allocator/allocator.c)                                                                                                                                                          
                                                            
  add_chunk and remove_chunk held the global mutex while performing GPU memory allocation/free (cuMemAlloc_v2 / cuMemoryFree). With 8 concurrent NCCL workers, heavy lock contention disrupted NCCL's internal state.
                                                            
  Fix: Move GPU allocation/free operations outside the mutex lock. Only pure in-memory tracking (linked list insert/remove + shared memory usage updates) remains inside the lock. Lock hold time drops from         
  milliseconds to microseconds.                             
                                                                                                                                                                                                                     
  2b. CUDA graph function interception (src/cuda/hook.c)                                                                                                                                                             
  
  HAMi-core's cuGetProcAddress hook returned pass-through wrappers for cuGraph* functions. Although these wrappers simply forwarded to the real CUDA driver, the extra indirection and LOG_DEBUG calls interfered    
  with NCCL's CUDA graph state management during capture.   
                                                                                                                                                                                                                     
  Fix: Return NULL for cuGraph* functions in find_symbols_in_table, letting them fall through to the real CUDA driver.                                                                                               
  
  Verification                                                                                                                                                                                                       
                                                            
  - vLLM 0.19.1 + Qwen/Qwen3-0.6B + tensor-parallel=8 + 8x RTX 3090                                                                                                                                                  
  - Before: guaranteed segfault with CUDA graphs enabled; only --enforce-eager worked
  - After: CUDA graph mode starts normally, graph capture completes in ~13s, inference works correctly

### DIFF
--- a/src/allocator/allocator.c
+++ b/src/allocator/allocator.c
@@ -89,7 +89,7 @@ CUresult get_listsize(allocated_list *al, size_t *size) {
 
 void allocator_init() {
     LOG_DEBUG("Allocator_init\n");
-    
+
     device_overallocated = malloc(sizeof(allocated_list));
     LIST_INIT(device_overallocated);
     device_allocasync=malloc(sizeof(allocated_list));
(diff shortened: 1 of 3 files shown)


archlitchi (approved):
/lgtm

PR #194, opened 2026-05-13
fix shrreg initialization failure handling

huazq:
## Summary
- Stop `try_create_shrreg` from continuing after shared-region file setup failures.
- Check `mmap` failures with `MAP_FAILED` and clean up partially initialized fd/mapping state before exiting.
- Use syscall-appropriate signed return types for `lseek` and `write` checks.

## Test Plan
- [x] `make check-cuda-hook-consistency`
- [x] `make build`
- [x] Temporary regression harness with invalid `CUDA_DEVICE_MEMORY_SHARED_CACHE` exits without SIGSEGV

### DIFF
--- a/src/multiprocess/multiprocess_memory_limit.c
+++ b/src/multiprocess/multiprocess_memory_limit.c
@@ -1090,28 +1090,34 @@ void try_create_shrreg() {
     int fd = open(shr_reg_file, O_RDWR | O_CREAT, 0666);
     if (fd == -1) {
         LOG_ERROR("Fail to open shrreg %s: errno=%d", shr_reg_file, errno);
+        goto fail;
     }
     region_info.fd = fd;
-    size_t offset = lseek(fd, SHARED_REGION_SIZE_MAGIC, SEEK_SET);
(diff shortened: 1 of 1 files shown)


src/multiprocess/multiprocess_memory_limit.c
```diff
@@ -1173,6 +1179,18 @@ void try_create_shrreg() {
...
+fail:
+    if (region_info.shared_region != NULL && region_info.shared_region != MAP_FAILED) {
+        munmap(region_info.shared_region, SHARED_REGION_SIZE_MAGIC);
+    }
+    region_info.shared_region = NULL;
+    if (fd != -1) {
+        close(fd);
+    }
+    region_info.fd = -1;
+    exit(-1);
```
DSFans2014: When pass -1 to exit(), it is interpreted as 255 . It is therefore recommended to use `EXIT_FAILURE` for better clarity and portability

huazq: thx for pointing it out; I have revised it according to your suggestions.

DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

Issue #148, opened 2026-01-08
initialization of gpu utilization for multiple gpus looks incorrect?


nishitnshah:
```
int init_gpu_device_utilization(){
    int i,dev;
    ensure_initialized();
    lock_shrreg();
    for (i=0;i<region_info.shared_region->proc_num;i++){
        for (dev=0;dev<CUDA_DEVICE_MAX_COUNT;dev++){
            region_info.shared_region->procs[i].device_util[dev].sm_util = 0;
            region_info.shared_region->procs[i].monitorused[dev] = 0;
            break;
        }
    }
    unlock_shrreg();
    return 1;
}
```
The break in the for loop looks buggy? It will only initialize the device 0 and proc 0. What about other procs and devices?

mesutoezdil:
Note: As of last week, this issue/PR has been flagged under our new activity policy. If there is no response or update by July 18, 2026, it will be closed. It can be reopened at any time if still relevant.

PR #199, opened 2026-05-29
 Fix SM utilization reporting for forked CUDA worker processes

lyquid617:
## Problem

  For Python multiprocessing workloads using HAMi-core, `Device_utilization_desc_of_container` may stay at 0 even when GPU processes are actively running.

  In the affected workload:

  - `nvidia-smi pmon` inside the container reports ~97% SM utilization per worker process.
  - `nvmlDeviceGetProcessUtilization` inside the container returns correct per-process samples.
  - `vGPU_device_memory_usage_in_bytes` is updated correctly.
  - `Device_utilization_desc_of_container` remains 0.
  - The shared cache has valid memory usage and updated `last_kernel_time`, but `device_util[].sm_util` stays 0.

  ## Root Cause

  `postInit()` starts the utilization watcher via `init_utilization_watcher()`, but it is guarded by `pthread_once(post_cuinit_flag)` and is only reliably triggered through the
`cuInit()` wrapper.

  For forked Python multiprocessing workers, the child process may inherit a completed `post_cuinit_flag` from the parent. Later CUDA kernel launches in the worker call the hooked launch path, but `postInit()` is skipped, so host PID detection and utilization watcher initialization never happen in the worker process.

  There is also an existing multi-GPU bug in `init_gpu_device_utilization()`: the inner loop breaks after device 0, so only the first device is reset. See[#148 ](https://github.com/Project-HAMi/HAMi-core/issues/148)

  ## Fix

  - Add `ensure_post_init()` and call it from kernel launch wrappers.
  - Reset `post_cuinit_flag` and `pidfound` in the child process after fork.
  - Remove the erroneous `break` in `init_gpu_device_utilization()`.

  ## Validation

  Tested with an 8-GPU Python multiprocessing workload:

  Before:

  - `nvidia-smi pmon`: ~97% SM utilization
  - `Device_utilization_desc_of_container`: 0

  After:

  - `Device_utilization_desc_of_container`: 97-98 per GPU
  - vGPU memory metrics remain correct

### DIFF
--- a/src/cuda/memory.c
+++ b/src/cuda/memory.c
@@ -3,6 +3,7 @@
 
 #include "allocator/allocator.h"
 #include "include/libcuda_hook.h"
+#include "include/libvgpu.h"
 #include "include/memory_limit.h"
 
 extern int pidfound;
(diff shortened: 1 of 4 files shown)


archlitchi:
thanks for the fix, please sign-off your commit

lyquid617:
> thanks for the fix, please sign-off your commit

fixed

archlitchi (approved):
/lgtm

PR #201, opened 2026-06-04
fix: prevent unsigned integer underflow in NVML memory info hook

### LINKED ISSUE #200: Bug: NVML/CUDA memory hooks report corrupted values when usage exceeds limit (integer underflow)
## Description

External monitoring tools (e.g. Prometheus DCGM, nvidia-smi) periodically show **abnormally large GPU memory values** for pods using HAMi-core. The reported `free` memory wraps around to a huge value (e.g. 18 EB) when tracked `usage` exceeds the configured `limit`.

### 1. NVML Hook: Integer Underflow in `_nvmlDeviceGetMemoryInfo`

**File:** `src/nvml/hook.c`

```c
} else {
    switch (version) {
    case 1:
        ((nvmlMemory_t*)memory)->free = (limit-usage);  // BUG: underflow when usage > limit
        ((nvmlMemory_t*)memory)->total = limit;
        ((nvmlMemory_t*)memory)->used = usage;
        return NVML_SUCCESS;
    case 2:
        ((nvmlMemory_v2_t *)memory)->free = (limit-usage);  // BUG: same issue
        ((nvmlMemory_v2_t *)memory)->total = limit;
        ((nvmlMemory_v2_t *)memory)->used = usage;
```

When `usage > limit` (due to stale shared memory slots from crashed processes, or race conditions), `limit - usage` produces a **massive `size_t` value** (e.g. `18446744073709551615`). External monitoring that calls `nvmlDeviceGetMemoryInfo` then sees this garbage value.

### 2. CUDA Hook: Returns Error Instead of Clamping

**File:** `src/cuda/memory.c`

```c
} else if (limit < usage) {
    LOG_WARN("limit < usage; usage=%ld, limit=%ld", usage, limit);
    return CUDA_ERROR_INVALID_VALUE;  // BUG: error instead of defensive clamping
}
```

When this happens inside the CUDA runtime path (`cuMemGetInfo_v2`), it returns an error code. Downstream callers may not handle this gracefully, leading to inconsistent state.

### Why `usage > limit` Occurs

The shared memory region (`/tmp/cudevshr.cache`) tracks per-process GPU memory usage. When a process crashes without cleanup, its memory slot is not freed, and the accumulated `usage` can exceed the `limit`.

## Steps to Reproduce

1. Deploy a pod with HAMi-core GPU memory limit (e.g., `volcano.sh/vgpu-memory: 4096`)
2. Run GPU workloads that crash or get OOM-killed
3. Monitor via external DCGM exporter or node-level `nvidia-smi`
4. Observe occasional spikes of `free` memory showing impossibly large values

### 1. Clamp NVML hook return values (src/nvml/hook.c)

```c
} else {
    // Defensive: clamp usage to limit
    size_t clamped_usage = (usage > limit) ? limit : usage;
    switch (version) {
    case 1:
        ((nvmlMemory_t*)memory)->free = limit - clamped_usage;
        ((nvmlMemory_t*)memory)->total = limit;
        ((nvmlMemory_t*)memory)->used = clamped_usage;
        return NVML_SUCCESS;
    case 2:
        ((nvmlMemory_v2_t *)memory)->free = limit - clamped_usage;
        ((nvmlMemory_v2_t *)memory)->total = limit;
        ((nvmlMemory_v2_t *)memory)->used = clamped_usage;
```

### 2. Clamp CUDA hook instead of returning error (src/cuda/memory.c)

```c
} else if (limit < usage) {
    LOG_WARN("limit < usage; usage=%ld, limit=%ld, clamping", usage, limit);
    // Clamp instead of returning error
    CUDA_OVERRIDE_CALL(cuda_library_entry, cuMemGetInfo_v2, free, total);
    size_t actual_limit = (limit > *total) ? *total : limit;
    *free = 0;
    *total = actual_limit;
    return CUDA_SUCCESS;
}
```

### 3. Consider periodic stale slot cleanup

Add a mechanism to detect and clean up process slots whose PID is no longer alive, to prevent `usage` from drifting above `limit` in the first place.

## Impact

- External GPU monitoring (Prometheus/Grafana dashboards) shows unusable data
- Alerting rules may trigger false alarms
- May cause confusion for cluster operators

## Environment

- HAMi-core branch: `volcano` (also affects `main`)
- Commit: b2bd3dc
- libvgpu.so compiled with `-DHOOK_NVML_ENABLE=1 -DHOOK_MEMINFO_ENABLE=1 -DMULTIPROCESS_LIMIT_ENABLE=1`

miaobyte:
## Summary
Fix unsigned integer underflow in `_nvmlDeviceGetMemoryInfo`: when GPU memory usage exceeds the configured limit, `(limit - usage)` with unsigned `size_t` wraps to ~18 EB, breaking monitoring tools.

## Changes
`src/nvml/hook.c`: compute `clamped = (usage > limit) ? limit : usage`, then derive both `used = clamped` and `free = limit - clamped`. This guarantees `free + used = total` always holds and `free` never underflows.

Fixes #200

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed reported CUDA memory values when actual usage exceeds the configured memory limit.
  * Fixed reported NVML memory values (used/free/total) to avoid underflow and inaccurate results in over-usage scenarios.
  * Added warnings to notify when memory usage exceeds the configured limit.

### DIFF
--- a/src/nvml/hook.c
+++ b/src/nvml/hook.c
@@ -353,16 +353,17 @@ nvmlReturn_t _nvmlDeviceGetMemoryInfo(nvmlDevice_t device,void* memory,int versi
             return NVML_SUCCESS;
         }
     } else {
+        size_t clamped = (usage > limit) ? limit : usage;
         switch (version) {
         case 1:
-             ((nvmlMemory_t*)memory)->free = (limit-usage);
+             ((nvmlMemory_t*)memory)->free = limit - clamped;
              ((nvmlMemory_t*)memory)->total = limit;
-             ((nvmlMemory_t*)memory)->used = usage;
+             ((nvmlMemory_t*)memory)->used = clamped;
             return NVML_SUCCESS;
         case 2:
-            ((nvmlMemory_v2_t *)memory)->free = (limit-usage);
+            ((nvmlMemory_v2_t *)memory)->free = limit - clamped;
             ((nvmlMemory_v2_t *)memory)->total = limit;
-            ((nvmlMemory_v2_t *)memory)->used = usage;
+            ((nvmlMemory_v2_t *)memory)->used = clamped;
             return NVML_SUCCESS;
         } 
     }
--- a/src/nvml/hook.c
+++ b/src/nvml/hook.c
@@ -358,17 +358,23 @@ nvmlReturn_t _nvmlDeviceGetMemoryInfo(nvmlDevice_t device,void* memory,int versi
             LOG_WARN("NVML meminfo: usage %lu exceeds limit %lu, clamping", usage, limit);
         }
         switch (version) {
-        case 1:
-             ((nvmlMemory_t*)memory)->free = limit - clamped;
-             ((nvmlMemory_t*)memory)->total = limit;
-             ((nvmlMemory_t*)memory)->used = clamped;
+        case 1: {
+            size_t physical_total = ((nvmlMemory_t*)memory)->total;
+            size_t actual_limit = (limit > physical_total) ? physical_total : limit;
+            ((nvmlMemory_t*)memory)->free = (actual_limit > clamped) ? (actual_limit - clamped) : 0;
+            ((nvmlMemory_t*)memory)->total = actual_limit;
+            ((nvmlMemory_t*)memory)->used = clamped;
             return NVML_SUCCESS;
-        case 2:
-            ((nvmlMemory_v2_t *)memory)->free = limit - clamped;
-            ((nvmlMemory_v2_t *)memory)->total = limit;
+        }
+        case 2: {
+            size_t physical_total = ((nvmlMemory_v2_t *)memory)->total;
+            size_t actual_limit = (limit > physical_total) ? physical_total : limit;
+            ((nvmlMemory_v2_t *)memory)->free = (actual_limit > clamped) ? (actual_limit - clamped) : 0;
+            ((nvmlMemory_v2_t *)memory)->total = actual_limit;
             ((nvmlMemory_v2_t *)memory)->used = clamped;
             return NVML_SUCCESS;
-        } 
+        }
+        }
     }
     return NVML_SUCCESS;
 }
--- a/src/nvml/hook.c
+++ b/src/nvml/hook.c
@@ -354,6 +354,9 @@ nvmlReturn_t _nvmlDeviceGetMemoryInfo(nvmlDevice_t device,void* memory,int versi
         }
     } else {
         size_t clamped = (usage > limit) ? limit : usage;
+        if (usage > limit) {
+            LOG_WARN("NVML meminfo: usage %lu exceeds limit %lu, clamping", usage, limit);
+        }
         switch (version) {
         case 1:
              ((nvmlMemory_t*)memory)->free = limit - clamped;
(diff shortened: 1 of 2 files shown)


src/nvml/hook.c
```diff
@@ -353,16 +353,17 @@ nvmlReturn_t _nvmlDeviceGetMemoryInfo(nvmlDevice_t device,void* memory,int versi
             return NVML_SUCCESS;
         }
     } else {
+        size_t clamped = (usage > limit) ? limit : usage;
         switch (version) {
         case 1:
-             ((nvmlMemory_t*)memory)->free = (limit-usage);
+             ((nvmlMemory_t*)memory)->free = limit - clamped;
```
mesutoezdil: This silently clamps the value without logging, so operators can't tell usage actually exceeded the limit.

mesutoezdil:
This only fixes the NVML hook, not the CUDA hook underflow from the same issue, so closing #200 here would be premature.

src/nvml/hook.c
maverick-woo: Agreed, a LOG_WARN here would help operators detect when usage exceeded the limit. Something like:

```c
size_t clamped = (usage > limit) ? limit : usage;
if (usage > limit) {
    LOG_WARN("NVML meminfo: usage %lu exceeds limit %lu, clamping", usage, limit);
}
```

maverick-woo:
已收到，正在 review 你的 PR。

maverick-woo:
Thanks for the review @mesutoezdil. Agreed on both points:

1. ✅ **Need LOG_WARN** — silently clamping hides the fact that usage drifted above limit, which operators need visibility into
2. ✅ **CUDA hook also needs fixing** — the `return CUDA_ERROR_INVALID_VALUE` path in `src/cuda/memory.c` should be similarly clamped rather than returning an error, otherwise applications calling `cuMemGetInfo` may crash or behave unpredictably

@miaobyte please update when you have a chance.

miaobyte:
@maverick123123  @mesutoezdil 

done,now updated:

**NVML hook (`src/nvml/hook.c`)**
- Added `LOG_WARN("NVML meminfo: usage %lu exceeds limit %lu, clamping", ...)` when `usage > limit`, so operators have visibility into the over-limit condition.

**CUDA hook (`src/cuda/memory.c`)**
- Removed the `return CUDA_ERROR_INVALID_VALUE` path when `usage > limit` — it now clamps usage to limit and reports `free=0`, `total=actual_limit`, matching the NVML hook behavior. This prevents `cuMemGetInfo` callers from crashing or seeing unpredictable values.
- Added `LOG_WARN("CUDA meminfo: usage %lu exceeds limit %lu, clamping", ...)` with the same pattern for consistency.

Both hooks now handle `usage > limit` identically: clamp, warn, and report `free=0` / `total=limit`.

maverick-woo:
收到，我们先测试一下。

maverick-woo:
测试通过，8x RTX 3090 上编译和运行均正常。@archlitchi 可以 merge 了。

mesutoezdil:
@miaobyte Thanks for the updates, LOG_WARN and the CUDA fix look good.

One item from CodeRabbit's review is still open though: on the NVML side, `total` is set to `limit` unconditionally, without capping it against the GPU's actual physical total like the CUDA path does (`actual_limit = min(limit, *total)`). If the configured limit is set higher than the real GPU memory, NVML would report an impossible total.

Can you apply the same min() clamp on the NVML side before we merge?

mesutoezdil:
pls remove claude commits from history

mesutoezdil:
and pls pls refrain from force pushing once you have started getting reviews

archlitchi (approved):
/lgtm

PR #202, opened 2026-06-10
remove duplicate --print arg check in shrreg_tool

mesutoezdil:
The --print case was checked twice in the arg loop inside shrreg_tool.c.
This caused print_shared_region() to run twice when --print was passed.

Removed the second duplicate block. No behavior change for other args.


DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

PR #203, opened 2026-06-11
build: upgrade compile image to CUDA 13.3.0 ubuntu24.04

spencercjh:
## Summary

Upgrade the compile base image from `nvidia/cuda:12.9.1-cudnn-devel-ubuntu20.04` to `nvidia/cuda:13.3.0-cudnn-devel-ubuntu24.04`.

## Changes

- **Makefile** & **Dockerfile**: bump base image to CUDA 13.3.0 + Ubuntu 24.04
- **src/multiprocess/multiprocess_utilization_watcher.c**: fix GCC 14 `-Wreturn-mismatch` (`return;` → `return NULL;` in `void*` function)
- **src/utils.c**: add missing `#include <sys/file.h>` for `flock()` to fix implicit declaration error under GCC 14
- **dockerfiles/Dockerfile**: add `git config --global --add safe.directory /libvgpu` to ensure `build.sh` can read git metadata during `docker build`

## Verification

- `make build-in-docker` builds successfully
- `docker build -f dockerfiles/Dockerfile .` builds successfully
- Both produce `libvgpu.so` and test binaries as expected

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ build-in-docker:
 		-v $(current_dir):/libvgpu \
 		-w /libvgpu \
 		-e DEBIAN_FRONTEND=noninteractive \
-		nvidia/cuda:12.9.1-cudnn-devel-ubuntu20.04 \
+		nvidia/cuda:13.3.0-cudnn-devel-ubuntu24.04 \
 		sh -c "apt-get -y update && \
            apt-get -y install cmake git && \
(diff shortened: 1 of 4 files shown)


spencercjh:
Part of https://github.com/Project-HAMi/HAMi/issues/1889

spencercjh:
WE have to use ubi8 based image.

PR #207, opened 2026-06-16
fix file descriptor leak in load_env_from_file

mesutoezdil:
`fopen` was called but `fclose` was never called before `return`, leaking a file descriptor on every invocation of `load_env_from_file`.


DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

PR #208, opened 2026-06-16
fix feof antipattern in load_env_from_file

mesutoezdil:
`feof()` only returns true after a read has already hit EOF, not before. The previous code called `fgets` unconditionally inside the loop without checking its return value, which could process stale buffer content on the last iteration.

Replace `while (!feof(f))` + bare `fgets` with `while (fgets(...) != NULL)` so the loop exits as soon as the read fails.

### DIFF
--- a/src/multiprocess/multiprocess_memory_limit.c
+++ b/src/multiprocess/multiprocess_memory_limit.c
@@ -154,17 +154,17 @@ int load_env_from_file(char *filename) {
     if (f==NULL)
         return 0;
     char tmp[10000];
-    int cursor=0;
-    size_t tmplen=0;
+    int cursor = 0;
+    size_t tmplen = 0;
     while (fgets(tmp, sizeof(tmp), f) != NULL) {
-        if (strstr(tmp,"=")==NULL)
+        if (strstr(tmp, "=") == NULL)
             break;
-        tmplen=strlen(tmp);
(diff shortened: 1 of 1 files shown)


DSFans2014 (commented):
/lgtm

src/multiprocess/multiprocess_memory_limit.c
DSFans2014: Additionally, it's better to cache the value of `strlen` to avoid repeated calls. If you're willing to make changes, please also revise this part accordingly.

mesutoezdil: done

src/multiprocess/multiprocess_memory_limit.c
```diff
@@ -155,13 +155,14 @@ int load_env_from_file(char *filename) {
         return 0;
     char tmp[10000];
     int cursor=0;
-    while (!feof(f)){
-        fgets(tmp,10000,f);
+    size_t tmplen=0;
```
DSFans2014: please resolve the lint error

mesutoezdil: done

DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

PR #209, opened 2026-06-16
cache getenv result for CUDA_TASK_PRIORITY_ENV

mesutoezdil:
`getenv` was called twice in sequence: once to check for NULL and once to read the value. Store the result in a local variable to avoid the redundant call.


DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

PR #210, opened 2026-06-16
check malloc return value in allocator_init

mesutoezdil:
`LIST_INIT` dereferences the pointer immediately after `malloc`. If `malloc` returns NULL (e.g. under memory pressure), the process crashes with a segfault and no diagnostic output.

Add a NULL check after each `malloc` call and exit with an error log so the failure is visible.


DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

PR #211, opened 2026-06-16
replace sprintf with snprintf in proc_alive

mesutoezdil:
`proc_alive` in `process_utils.h` still used `sprintf`. Replace with `snprintf` to be consistent with the rest of the codebase.


DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

PR #212, opened 2026-06-18
remove enable_eager_execution call in TF2 test

mesutoezdil:
`tf.enable_eager_execution()` raises `ValueError` in TF2 since eager mode is on by default. Removed the call.


archlitchi (approved):
/lgtm

PR #213, opened 2026-06-19
fix: pass device arg to cuda() in pytorch test

mesutoezdil:
`--device` argument was parsed but never forwarded to `.cuda()`, so the test always ran on GPU 0 regardless of input.

Fix: `.cuda()` -> `.cuda(args.device)`, consistent with `limit_mxnet.py` and `limit_tensorflow.py`.


archlitchi (approved):
/lgtm

Issue #67, opened 2025-04-17
use cuGraphAddMemAllocNode to create mem, but free error


huanghuangzym:
we run  cuda-samples in hami
the code is  
https://github.com/NVIDIA/cuda-samples/blob/v12.4.1/Samples/3_CUDA_Features/graphMemoryFootprint/graphMemoryFootprint.cu

the  unfreedAllocations case ,when free will failed
------------
this is run log

================================
Running unfreed streams example.
CUDA cannot reuse phyiscal memory from graphs which do not free their allocations.

[HAMI-core Info(27:140685756022784:stream.c:4)]: cuStreamCreate 0x7fffa0dd9248
Despite being launched in the same stream, each graph launch grows the memory footprint. Since the allocation is not freed, CUDA keeps the memory valid for use.
00:     FOOTPRINT: 67108864 bytes
01:     FOOTPRINT: 134217728 bytes
02:     FOOTPRINT: 201326592 bytes
03:     FOOTPRINT: 268435456 bytes
04:     FOOTPRINT: 335544320 bytes
05:     FOOTPRINT: 402653184 bytes
06:     FOOTPRINT: 469762048 bytes
07:     FOOTPRINT: 536870912 bytes

Trimming does not impact the memory footprint since the un-freed allocations are still holding onto the memory.
    FOOTPRINT: 536870912 bytes
[HAMI-core ERROR (pid:27 thread=140685756022784 allocator.c:166)]: remove_chunk  a_list length is 0 res=0
[HAMI-core Info(27:140685756022784:memory.c:200)]: after free_raw dptr=0x320000000 res=-1
CUDA error at graphMemoryFootprint.cu:344 code=-1(unrecognized error code) "cudaFree(dPtrs[i])" 
[HAMI-core Msg(27:140685756022784:multiprocess_memory_limit.c:499)]: Calling exit handler 27

it seems ,hami not hook cuGraphAddMemAllocNode ?  but user can free by ptr, then it free error ?

mesutoezdil:
Hi,

We are going through open issues to keep things manageable. This issue was opened before 2026 and has not been resolved or updated recently.

Please reply by 2026-07-13 to confirm this is still relevant and based on the current codebase. If we do not hear back, we will close this issue. You are welcome to reopen it or submit a new one when ready.

Thank you for your contribution to HAMi.

PR #217, opened 2026-07-07
Track cuMemAllocFromPoolAsync allocations and free untracked async pointers

### LINKED ISSUE #93: Implement similar logic to interface cuMemAllocAsync for interface cuMemAllocFromPoolAsync
when we use cuMemAllocFromPoolAsync to allocate and use cuMemFreeAsync to free  in  the sparkrapid  ASYNC rmm pool usage scenario,it will coredump ai.rapids.cudf.CudfException: transform: failed inside CUB: cudaErrorInvalidDevice: invalid device ordinal .

imlach:
Fixes #93. Relates to #67.

## The bug

libvgpu tracks stream-ordered allocations so it can enforce the memory limit and
free them correctly. It tracks the ones from `cuMemAllocAsync`, but
`cuMemAllocFromPoolAsync` was left as a bare pass-through, so a pool allocation
was never registered.

When that pointer is freed, `remove_chunk_async` walks the tracking list, fails
to find it, and returns -1 without calling the real `cuMemFreeAsync`. The caller
gets -1, which isn't a valid `CUresult` and shows up as "unrecognized error
code" (issue #93, e.g. a RAPIDS RMM coredump), and the memory leaks.

We ran into this getting NVIDIA Dynamo's snapshot feature working on a
HAMi-managed GPU, and reduced it to the minimal reproducer in the test.

## The fix

`cuMemAllocFromPoolAsync` now registers its allocation the same way
`cuMemAllocAsync` does, so the matching free finds and releases it. It is
accounted against the caller's pool (the one it allocates from) rather than the
device default pool; a per-pool limit would be the alternative if you'd prefer
it. The shared registration and accounting logic is factored into one helper so
the two paths can't drift.

`remove_chunk_async` also stops returning -1 for pointers it doesn't recognise:
it now frees them for real and returns the driver's own result. Tracked pointers
are unaffected, since they return earlier in the match loop, so there is no
double-free.

Two smaller changes fell out of this: both async alloc paths now return
`CUDA_ERROR_OUT_OF_MEMORY` instead of -1 when over the limit (RMM keys its
cache-release-and-retry on that code), and the shared limit counter can no
longer underflow when a pool reports no reserved high-water.

## Known limitations

- Under a mix of default-pool and explicit-pool allocations, the single shared
  limit counter can under-account a pool allocation (it may register as zero), so
  enforcement is approximate for multi-pool workloads. Still better than before,
  when pool allocations were not tracked at all; per-pool accounting is the
  follow-up.
- A pointer allocated synchronously (`cuMemAlloc`) but freed with `cuMemFreeAsync`
  is now freed correctly, but its synchronous tracking entry is not reclaimed, so
  it stays accounted. This is unusual, and a proper fix needs the synchronous
  free path, which this PR leaves alone.
- Accounting uses the current device, so an explicit pool living on another
  device is charged to the wrong limit. Benign in single-GPU containers and when
  the caller selects the device first (as RMM does).

## Testing

Adds `test/test_alloc_pool_async.c`, picked up automatically by the test CMake
glob. It exercises two paths on one stream: a default-pool control
(`cuMemAllocAsync`) and the pool-explicit case. On stock CUDA both free cleanly;
under libvgpu the pool-explicit free fails before this change and succeeds after.

Built this branch's `libvgpu.so` and ran it on an Ampere GPU under a HAMi v2.9.0
deployment: the pool-explicit free goes from -1 ("unrecognized error code") to
`CUDA_SUCCESS`, with the control path unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous allocation support from a specified device memory pool.
* **Bug Fixes**
  * Improved async allocation/free bookkeeping to be memory-pool aware.
  * Async frees now correctly handle previously untracked pointers and better respect pool high-water limits.
* **Tests**
  * Added a standalone reproducer verifying `cuMemFreeAsync` behavior for both default-pool and explicit pool async allocations.

### DIFF
--- a/src/allocator/allocator.c
+++ b/src/allocator/allocator.c
@@ -237,9 +237,6 @@ int free_raw(CUdeviceptr dptr) {
 int remove_chunk_async(
     allocated_list *a_list, CUdeviceptr dptr, CUstream hStream) {
     size_t t_size;
-    if (a_list->length == 0) {
-        return -1;
-    }
     allocated_list_entry *val;
     for (val = a_list->head; val != NULL; val = val->next) {
         if (val->entry->address == dptr) {
@@ -253,7 +250,9 @@ int remove_chunk_async(
             return 0;
         }
     }
-    return -1;
+    /* Not tracked by libvgpu: free it for real instead of returning -1, which
+     * leaked and surfaced as an "unrecognized error code". */
+    return CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemFreeAsync,dptr,hStream);
 }
 
 int free_raw_async(CUdeviceptr dptr, CUstream hStream) {
@@ -263,14 +262,38 @@ int free_raw_async(CUdeviceptr dptr, CUstream hStream) {
     return tmp;
 }
 
+/* Track an allocated async chunk and account it against its pool's high-water
+ * limit. Call with mutex held. */
+static int account_async_chunk(allocated_list_entry *e, size_t size, CUmemoryPool pool) {
+    size_t poollimit;
+    CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemPoolGetAttribute,pool,CU_MEMPOOL_ATTR_RESERVED_MEM_HIGH,&poollimit);
+    if (res != CUDA_SUCCESS) {
+        LOG_ERROR("cuMemPoolGetAttribute failed res=%d",res);
+        return res;
+    }
+    if (poollimit != 0) {
+        if (poollimit> device_allocasync->limit) {
+            size_t allocsize = (poollimit-device_allocasync->limit < size)? poollimit-device_allocasync->limit : size;
+            add_gpu_device_memory_usage(getpid(), e->entry->dev, allocsize, 2);
+            device_allocasync->limit=device_allocasync->limit+allocsize;
+            e->entry->length=allocsize;
+        }else{
+            e->entry->length=0;
+        }
+    }else{
+        e->entry->length=0;
+    }
+    LIST_ADD(device_allocasync,e);
+    return 0;
+}
+
 int add_chunk_async(CUdeviceptr *address, size_t size, CUstream hStream) {
     size_t addr=0;
-    size_t allocsize;
     CUresult res = CUDA_SUCCESS;
     CUdevice dev;
     cuCtxGetDevice(&dev);
     if (oom_check(dev,size))
-        return -1;
+        return CUDA_ERROR_OUT_OF_MEMORY;
 
     allocated_list_entry *e;
     INIT_ALLOCATED_LIST_ENTRY(e, addr, size, dev);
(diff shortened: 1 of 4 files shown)


imlach:
**Per-pool accounting.** This is already called out under "Known
limitations": the shared `device_allocasync->limit` is deliberately reused (as
the existing default-pool path does), so a pool allocation can under-account
under mixed default/explicit-pool workloads. Per-pool tracking is the follow-up.
This PR's scope is just making pool allocations tracked and freeable at all,
which they weren't before.

**Leak-on-failure.** Real, but a pre-existing, file-wide pattern rather than
something new here. Entries are only ever released via `LIST_REMOVE`, which frees
all three allocations (`entry->allocHandle`, `entry`, then the entry itself).
Every `add_*` path that returns before `LIST_ADD` leaks the same way today:
`add_chunk_async` on `main` already does, on the same `cuMemAllocAsync` /
`cuDeviceGetMemPool` / `cuMemPoolGetAttribute` failure branches, and the new
`add_chunk_from_pool_async` mirrors that convention.

Two notes on the suggested patch:

- `free(e)` alone would still leak `e->entry` and `e->entry->allocHandle`, since
  `INIT_ALLOCATED_LIST_ENTRY` mallocs all three; the correct cleanup is the same
  triple-free `LIST_REMOVE` performs.
- The `cuDeviceGetMemPool` device-leak branch in `add_chunk_async` is unchanged
  from `main`.

I'd rather fix these error paths consistently across the async allocator (the
correct triple-free on every early return, and deferring `*address` until setup
succeeds) in a focused follow-up than piecemeal in one function here. Happy to do
that; let me know if you'd prefer it land in this PR.

archlitchi:
Thanks, we'll validate this PR on our environment

maverick-woo:
@archlitchi 
 Tested this branch end-to-end on real hardware — works as intended.

  **Environment**

  | Item | Value |
  |---|---|
  | GPU / driver | 8× RTX 3090, driver 580.126.20 (CUDA 13.0) |
  | Build image | `nvcr.io/nvidia/cuda:13.0.0-devel-ubuntu24.04`, `build.sh` default flags |
  | Run | single GPU, `LD_PRELOAD=libvgpu.so`, `CUDA_DEVICE_MEMORY_LIMIT` set |

  **Results**

  | # | Test | Case | Result | Evidence |
  |---|---|---|---|---|
  | 1 | Reproducer `test_alloc_pool_async` | baseline, no preload | ✅ PASS | both async-free paths accepted (test valid on HW) |
  | 2 | Reproducer `test_alloc_pool_async` | under libvgpu, 2g | ✅ PASS | `[B] cuMemFreeAsync -> 0 (CUDA_SUCCESS)`; new path runs (`Hijacking cuMemAllocFromPoolAsync`) |
  | 3 | Accounting stress, 1 pool | 200× `alloc(32MiB)+free` @ 512 MiB | ✅ no leak/drift | would OOM by iter 16 if free didn't deaccount |
  | 4 | Accounting stress, 1 pool | alloc-without-free @ 512 MiB | ✅ limit enforced | OOM at 512 MiB with `CUDA_ERROR_OUT_OF_MEMORY` (not -1); log `Device 0 OOM 570425344 / 536870912` |
  | 5 | RAPIDS RMM (`CudaAsyncMemoryResource`, rmm-cu13 26.06) | baseline | ✅ OK | alloc / free / reuse all fine |
  | 6 | RAPIDS RMM | under libvgpu 2g: alloc 4×64 MiB + free | ✅ OK | the exact free that failed pre-fix (#93) |
  | 7 | RAPIDS RMM | under libvgpu 2g: reuse ×50 | ✅ no drift | — |
  | 8 | RAPIDS RMM | under libvgpu 2g: grow → OOM | ✅ clean | `MemoryError: std::bad_alloc: out_of_memory`, no crash |

  **RMM hook activity (proves the new path actually executed)**

  | Counter | Value | Meaning |
  |---|---|---|
  | `Hijacking cuMemAllocFromPoolAsync` | 83 | RMM drove the PR's new `add_chunk_from_pool_async` 83× |
  | `Hijacking cuMemFreeAsync` | 82 | diff of 1 = the single OOM-rejected alloc |
  | `Device 0 OOM` | 1 | limit enforced exactly once |

archlitchi (approved):
/lgtm

Issue #11, opened 2024-08-01
Error when sh build.sh


z19311:
Environment
Ubuntu 18.04 + CUDA 12.4

Error occurs when excuting **sh build.sh**

build.sh: 7: build.sh: [[: not found
build.sh: 11: build.sh: CMAKE_OPTIONS+= -DDLSYM_HOOK_ENABLE=1: not found
build.sh: 13: build.sh: CMAKE_OPTIONS+= -DMULTIPROCESS_LIMIT_ENABLE=1: not found
build.sh: 14: build.sh: CMAKE_OPTIONS+= -DHOOK_MEMINFO_ENABLE=1: not found
build.sh: 15: build.sh: CMAKE_OPTIONS+= -DHOOK_NVML_ENABLE=1: not found
build.sh: 16: build.sh: CMAKE_OPTIONS+= -DCMAKE_BUILD_TYPE=Debug: not found
CI_COMMIT_BRANCH:unknown
CI_COMMIT_SHA:fb11ec3d83318c4c7924c1675071d1cc778a7472
-- Git hash is fb11ec3
CMake Error at src/multiprocess/CMakeLists.txt:4 (target_link_libraries):
  Object library target "multiprocess_mod" may not link to anything.

-- Configuring incomplete, errors occurred!
See also "/home/HAMi-core/build/CMakeFiles/CMakeOutput.log".
make: *** No targets specified and no makefile found.  Stop.

Could you show us the build environment?

daiqizhi:
bash build.sh or chmod +x build.sh ; ./build.sh

jiangxiaobin96:
> bash build.sh or chmod +x build.sh ; ./build.sh

Same problem. `bash build.sh or chmod +x build.sh ; ./build.sh` doesn't work.

PR #219, opened 2026-07-10
Support Kai-resource-isolator monitoring

archlitchi:
we need to construct the right cache file from multiple downward envs from kai-resource-isolator and kai-scheduler

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shared-region cache path selection when no explicit cache path is configured.
  * When runtime container identifiers are available, the system now uses a container-scoped cache location to avoid collisions.
  * If the synthesized cache location can’t be used, it automatically falls back to the standard default cache location.


src/multiprocess/multiprocess_memory_limit.c
```diff
@@ -1075,8 +1075,30 @@ void try_create_shrreg() {
...
+        const char* container_name = getenv(MULTIPROCESS_SHARED_REGION_CONTAINER_NAME_ENV);
+
+        if (hook_path == NULL || hook_path[0] == '\0') {
+            hook_path = MULTIPROCESS_SHARED_REGION_HOOK_PATH_DEFAULT;
+        }
+
+        if (pod_uid != NULL && pod_uid[0] != '\0' &&
+            container_name != NULL && container_name[0] != '\0') {
+            snprintf(synthesized_cache_path, sizeof(synthesized_cache_path),
+                     "%s/containers/%s_%s",
```
dttung2905: this opens `containers/{uid}_{name}` as the mmap file, but vGPUmonitor only scans directories under `containers/` and looks for a `*.cache` file inside (`cudevshr.go`). HAMi device plugin also uses `containers/{uid}_{name}/` as a directory. Could libvgpu use `containers/{uid}_{name}/usage.cache` instead?

mesutoezdil: Does `open()` even survive if `containers/` doesn't already exist, or are we crashing before vgpumonitor gets a chance to miss the file?

dttung2905: Good catch. `open(..., O_CREAT)` does not create the parent directory. If `containers/` is missing, `open()` fails with `ENOENT`, and the `fail` path stops the workload.

The webhook creates and mounts this directory. However, HAMi-core should handle a missing directory safely. It should log a warning and use the default cache path instead of stopping the workload.

On the path format: kai-vgpu-monitor will follow this PR and read the flat file at `containers/{uid}_{name}`. https://github.com/Project-HAMi/KAI-resource-isolator/pull/11

archlitchi: > this opens `containers/{uid}_{name}` as the mmap file, but vGPUmonitor only scans directories under `containers/` and looks for a `*.cache` file inside (`cudevshr.go`). HAMi device plugin also uses `containers/{uid}_{name}/` as a directory. Could libvgpu use `containers/{uid}_{name}/usage.cache` instead?

it make sense, i'll swith to your design, thanks for pointing that out.

archlitchi: > Good catch. `open(..., O_CREAT)` does not create the parent directory. If `containers/` is missing, `open()` fails with `ENOENT`, and the `fail` path stops the workload.
> 
> The webhook creates and mounts this directory. However, HAMi-core should handle a missing directory safely. It should log a warning and use the default cache path instead of stopping the workload.
> 
> On the path format: kai-vgpu-monitor will follow this PR and read the flat file at `containers/{uid}_{name}`. [Project-HAMi/KAI-resource-isolator#11](https://github.com/Project-HAMi/KAI-resource-isolator/pull/11)

fixed

archlitchi:
@dttung2905 please take a look

src/multiprocess/multiprocess_memory_limit.c
```diff
@@ -1075,8 +1075,54 @@ void try_create_shrreg() {
...
+
+        if (pod_uid != NULL && pod_uid[0] != '\0' &&
+            container_name != NULL && container_name[0] != '\0') {
+            int dir_written = snprintf(synthesized_container_dir, sizeof(synthesized_container_dir),
+                                        "%s/containers/%s_%s",
+                                        hook_path, pod_uid, container_name);
+            int file_written = -1;
+            if (dir_written >= 0 && (size_t)dir_written < sizeof(synthesized_container_dir)) {
+                file_written = snprintf(synthesized_cache_path, sizeof(synthesized_cache_path),
+                                         "%s/usage.cache", synthesized_container_dir);
```
mesutoezdil, line 1098: kai-resource-isolator #11 skips dirs and reads the flat containers/{uid}_{name} file.. 
so this dir + usage.cache layout makes the monitor see nth unless #11 is updated too

archlitchi: yeah, i'll notify him

PR #223, opened 2026-07-17
fix(nvml): ensure nvml_postInit only runs on NVML_SUCCESS

fishman:
I closed this by mistake, while cleaning up forks. Old cleanup of https://github.com/Project-HAMi/HAMi-core/pull/168

Fixes a critical edge case where GPU initialization permanently fails in Kubernetes 1.27+ Burstable pods (e.g., when device files are temporarily locked during resize: InProgress).
Issue Related : https://github.com/Project-HAMi/HAMi-core/issues/167 and https://github.com/Project-HAMi/HAMi/issues/1704#issuecomment-4115925203

The Bug:
Because nvml_postInit used pthread_once, it was marked as "completed" even if the initial nvmlInit driver call failed. If an application retried the initialization later, nvml_postInit was bypassed, leaving the virtual GPU map permanently broken.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved initialization handling so follow-up setup runs only after successful NVML initialization.
  * Preserved initialization results and behavior when initialization fails.


archlitchi (approved):
/lgtm

PR #226, opened 2026-07-22
fix: avoid div by zero and underflow in pitch alloc size guess

mesutoezdil:
Zero WidthInBytes underflows the size guess to SIZE_MAX, zero ElementSizeBytes divides by zero and crashes. This guards both to 0 before the guess is computed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of pitched memory allocation requests when the element size or width is zero.
  * Prevented invalid pitch calculations, ensuring memory sizing and out-of-memory tracking remain consistent for these edge cases.


archlitchi (approved):
/lgtm

PR #227, opened 2026-07-22
fix: lock shared region before clearing proc slots in oom_check

mesutoezdil:
oom_check called clear_proc_slot_nolock without holding the shrreg lock, unlike its other caller. Concurrent OOM across processes can race on the shared proc table. Now it locks around the call, same as init_proc_slot_withlock.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved out-of-memory handling during resource allocation to safely retry after clearing an available process slot.

* **Documentation**
  * Clarified internal comments related to GPU allocation and resource tracking.


archlitchi (approved):
/lgtm

PR #230, opened 2026-07-23
fix: probe CUDA context on correct device in set_task_pid()

### LINKED ISSUE #225: set_task_pid: deterministic "host pid is error" when CUDA_VISIBLE_DEVICES does not start with the pod's device 0 (probe context and poll target mismatch), leading to spurious Device 0 OOM
## Summary

In a container with multiple vGPUs, `set_task_pid()` fails deterministically ("host pid is error!") for any process whose `CUDA_VISIBLE_DEVICES` does not **start with** the container's device 0. The failure is not a race: it reproduces 100% with a single process and no concurrency. After the failure, the process's memory usage is merged into the wrong accounting slot, so allocations that are legal per device get summed against one device's limit and trip a spurious `Device 0 OOM`.

This looks like the root cause behind reports such as Project-HAMi/HAMi#1455 (same log signature during multi-GPU training, closed as stale without analysis).

## Environment

- HAMi v2.7.1 (bundled libvgpu), also traced against HAMi-core `main` sources
- Kubernetes + hami-scheduler, pod with `nvidia.com/gpu: 2`, `nvidia.com/gpumem: 5800` (per device)
- NVIDIA driver 580.82.07, RTX 2080 Ti x2 (originally observed on A100-80GB x8, driver 580.105, verl FSDP world=8)
- PyTorch 2.x (`vllm/vllm-openai:v0.10.2` image), CUDA 12.x

## Reproduction

Inside a single pod that was allocated 2 vGPUs, run one Python process per case (no concurrency at all), each allocating 4096 MiB in 256 MiB steps (well under the 5800 MiB per-device cap):

```bash
CUDA_VISIBLE_DEVICES=<case> python3 -c '
import sys, torch
# optional: torch.cuda.set_device(1)
torch.cuda.init()
bufs = [torch.empty(256 << 20, dtype=torch.uint8, device="cuda") for _ in range(16)]
'
```

Results (3/3 identical across repetitions):

| case | CUDA_VISIBLE_DEVICES | context created on | result |
|---|---|---|---|
| A | `0` | pod dev 0 | ok |
| B | `1` | pod dev 1 | **host pid is error** |
| C | `1,0` | pod dev 1 (default) | **host pid is error** |
| D | `1,0` + `set_device(1)` | pod dev 0 | **host pid is error** |
| E | `0,1` + `set_device(1)` | pod dev 1 | ok |
| F | `0,1` | pod dev 0 | ok |

So the failure condition is exactly: **`CUDA_VISIBLE_DEVICES[0]` != the pod's first device**. Where the user code later places its context is irrelevant (C vs D vs E), because the probe happens during library initialization.

With two workers (CVD=`0` and CVD=`1`, e.g. Ray actors in one pod, or any launcher that narrows CVD per rank), the CVD=`1` worker fails host-pid detection every time, and its allocations are accounted together with the other worker's, producing `Device 0 OOM <sum>/<cap>` while each device is individually within its cap. Observed signatures: `Device 0 OOM 6337593344 / 6081740800` (2080 Ti), `88.79GB / 80GB` (A100 x8).

## Root cause (from source)

In `src/utils.c`, `set_task_pid()`:

1. The before/after process snapshots are taken from the **first CVD-visible device in NVML (physical) enumeration order** — the loop does an unconditional `break` after the first device where `nvml_to_cuda_map(i) >= 0`. For CVD=`1,0` or `0,1` that is the pod's physical device 0.
2. The probe context, however, is created with `cuDevicePrimaryCtxRetain(&pctx, 0)` — CUDA device 0, i.e. **whatever CVD lists first**.

When CVD=`1,0`, the poll watches physical device 0 while the probe context appears on physical device 1, so `getextrapid()` never sees a new PID and returns 0 -> `LOG_ERROR("host pid is error!")`.

When CVD=`1`, physical device 0 is not visible at all, `nvml_to_cuda_map(0)` returns -1 for every polled index that could match, no snapshot is taken (`previous == running == 0`), and `getextrapid()` again returns 0. Same error, different path.

All six cases above, plus the original multi-worker OOM, are consistent with this model.

## Suggested fix

Create the probe context on the same device that is being polled, e.g. retain the primary context for the `cudaDev` found by the map inside the loop instead of hardcoding device 0:

```c
cudaDev = nvml_to_cuda_map(i);
if (cudaDev < 0) continue;
/* poll nvml index i ... */
CHECK_CU_RESULT(cuDevicePrimaryCtxRetain(&pctx, cudaDev));
```

(or equivalently, poll the NVML device that corresponds to CUDA device 0). Either direction removes the mismatch and also fixes the CVD=`1` path, since the polled device and the probe device are then guaranteed to be the same.

## Workarounds we validated

- Keep the pod's first device first in CVD (e.g. `0,1` is safe, `1,0` is not), or
- unset CUDA_VISIBLE_DEVICES in the container and select devices with `torch.cuda.set_device()` per process, or
- use a 1-GPU-per-pod layout so every process's CVD starts with that pod's device 0.

Happy to provide the full reproduction manifests (Kubernetes pod specs) if useful.

maverick-woo:
## Problem

`set_task_pid()` in `src/utils.c` polls NVML processes on the first physical device where `nvml_to_cuda_map(i) >= 0`, but the probe context is always created on hardcoded CUDA device 0 via `cuDevicePrimaryCtxRetain(&pctx, 0)`.

When `CUDA_VISIBLE_DEVICES` does not start with the pod's device 0 (e.g., `CVD=1` or `CVD=1,0`), the polled NVML device and the probe context device are different physical GPUs. `getextrapid()` never finds the new PID, causing a spurious "host pid is error!" and subsequent fake Device 0 OOM due to incorrect memory accounting.

## Fix

Save the `cudaDev` found in the NVML enumeration loop and use it for both `cuDevicePrimaryCtxRetain` and `cuDevicePrimaryCtxRelease`, ensuring the probe context runs on the same device being polled.

Also guard against the case where no NVML device maps to a CUDA device (probeDev remains -1) by returning early before retaining any context.

## Testing

Verified on an 8x RTX 3090 machine with all CVD combinations:

| CUDA_VISIBLE_DEVICES | Before fix | After fix |
|---------------------|------------|-----------|
| `0` | OK | OK |
| `1` | host pid is error | OK |
| `0,1` | OK | OK |
| `1,0` | host pid is error | OK |

## Related

Fixes #225

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -101,7 +101,7 @@ nvmlReturn_t set_task_pid() {
     nvmlProcessInfo_t1 pre_pids_on_device[SHARED_REGION_MAX_PROCESS_NUM];
     nvmlProcessInfo_t1 pids_on_device[SHARED_REGION_MAX_PROCESS_NUM];
     nvmlDevice_t device;
-    nvmlReturn_t res;
+    nvmlReturn_t res = NVML_SUCCESS;
     CUcontext pctx;
     int i;
     CHECK_NVML_API(nvmlInit());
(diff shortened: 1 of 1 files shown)


maverick-woo:
@rootsongjc ready to merge

src/utils.c
```diff
@@ -111,26 +111,34 @@ nvmlReturn_t set_task_pid() {
...
-        }while(res==NVML_ERROR_INSUFFICIENT_SIZE); 
-        mergepid(&previous,&merged_num,(nvmlProcessInfo_t1 *)tmp_pids_on_device,pre_pids_on_device);
+        } while (res == NVML_ERROR_INSUFFICIENT_SIZE);
+        mergepid(&previous, &merged_num, (nvmlProcessInfo_t1 *)tmp_pids_on_device, pre_pids_on_device);
+        probeDev = cudaDev;
         break;
     }
+
+    if (probeDev < 0) {
+        LOG_ERROR("host pid is error!");
```
mesutoezdil: reusing the same log msg and error code as the hostpid==0 case below makes it hard to tell which one actually failed when debugging

src/utils.c
```diff
@@ -111,26 +111,34 @@ nvmlReturn_t set_task_pid() {
...
+        return NVML_ERROR_DRIVER_NOT_LOADED;
+    }
+
     previous = merged_num;
     merged_num = 0;
-    memset(tmp_pids_on_device,0,sizeof(nvmlProcessInfo_v1_t)*SHARED_REGION_MAX_PROCESS_NUM);
-    CHECK_CU_RESULT(cuDevicePrimaryCtxRetain(&pctx,0));
+    memset(tmp_pids_on_device, 0, sizeof(nvmlProcessInfo_v1_t) * SHARED_REGION_MAX_PROCESS_NUM);
+    CHECK_CU_RESULT(cuDevicePrimaryCtxRetain(&pctx, probeDev));
```
mesutoezdil, line 141: if sth in this loop fails after the ctx retain above it returns early and never releases the ctx so probeDev leaks

src/utils.c
maverick-woo: Good point, updated the log message to "No mapped CUDA device found for host PID detection!" to distinguish it from the hostpid==0 case below.

src/utils.c
maverick-woo: Agreed, fixed. Replaced the CHECK_NVML_API macro and the direct return in the second loop with explicit checks that go through a cleanup label, so cuDevicePrimaryCtxRelease is always called before returning.

iemAnshuman:
@maverick123123 ran the four CVD cases against #230 on 8×A100, driver 580.65.06, cold and with a warm holder. leak is gone, base leaves the probe context on GPU 1 for CVD=1 and 1,0, patched leaves nothing in all eight runs. but those same two still hit SET_TASK_PID FAILED, and the log says current processes num = 0 0 so nvml isnt seeing the probe on the device it polls at all, which I dont have an explanation for yet. tried CUDA_DEVICE_ORDER=PCI_BUS_ID in case CUDA and NVML disagreed on numbering, no change. you had all four passing on 3090 so maybe its this a100 or the 580 driver. not blocking, the leak fix stands on its own. logs and per case nvidia smi: https://gist.github.com/iemAnshuman/c3a0f5d8efc8871bc94816da1e7dcdd8

iemAnshuman:
@maverick123123 followed this up with a small standalone test of just the set_task_pid() steps, on four A100. The 'current processes num = 0 0' I reported above does not happen there. 'cuDeviceGetUuid()' on the retained device and 'nvmlDeviceGetUUID()' on the polled handle give the same UUID on all nine layouts I tried, '0' '1' '0,1' '1,0' '2' '3' '2,3' and again with another process on the gpu, and the host pid turns up 0.062 to 0.071ms after the retain

So on that node, the device is right, and nvml publishes far too fast for the 330 s postinit budget to come into it.

I cant close the gap yet though. That node is driver 580.82.07 on A100-PCIE-40GB, where the failing runs were 580.65.06 on 8xA100, and the test runs the nvml steps on their own with no libvgpu.so. Three things differ her, which is why I can't say whether 'CVD=1' passing here is down to the test skipping libvgpu.so or to the different card and driver. Next thing I would try is the same probe under libvgpu.so on the box that failed.

logs and checksums: https://gist.github.com/iemAnshuman/e062b7297464fd07edefa9c350a72395

maverick-woo:
Thanks for the thorough testing across the A100 setups — really appreciate the detailed breakdown. Glad the leak fix holds across all cases. I'll keep an eye on the remaining `SET_TASK_PID FAILED` issue you flagged with `CVD=1` / `1,0` on 580.65.06. Let me know if you find anything when testing with libvgpu.so on the failing node.

mesutoezdil:

PR #235, opened 2026-07-27
feat: fallback to /proc/1/environ for missing environment variables

### LINKED ISSUE #2125: [LFX Mentorship] Strengthen GPU memory isolation for child and SSH-started processes
## Summary

Investigate and strengthen HAMi GPU memory isolation when workloads create child processes or when users start processes from a new SSH/login shell that does not inherit the environment variables injected into the original container process.

This issue is the upstream tracking issue for a proposed CNCF LFX Mentorship program. The original report in #1112 contains useful analysis, but it was closed during stale-issue cleanup without a confirmed technical resolution. The first milestone here is therefore to reproduce the behavior on a current supported HAMi release before selecting an implementation.

## Problem statement

HAMi currently relies on runtime configuration being available to the process loading HAMi-core. In some child-process and SSH-started-process scenarios, the expected configuration may not be inherited or may be user-modifiable. This can cause inconsistent enforcement and makes the isolation boundary difficult to reason about.

## Scope

- Reproduce the reported behavior on a current supported HAMi release with documented hardware, driver, container runtime, and Kubernetes versions.
- Define the intended isolation and trust model with maintainers.
- Trace how HAMi configuration reaches the original process, child processes, and SSH/login shells.
- Evaluate configuration-delivery approaches that do not depend solely on mutable inherited environment variables.
- Implement the mentor-approved design with backward compatibility where practical.
- Add regression tests and user-facing troubleshooting/migration documentation.

## Expected deliverables

- A reproducible test case and root-cause analysis.
- A short design document covering security boundaries, compatibility, and failure modes.
- An implementation accepted by HAMi maintainers, or a validated design/prototype if kernel or driver constraints prevent a production implementation within the term.
- Automated tests covering the original process, child processes, and SSH/new-shell behavior.
- Updated documentation for operators and users.

## Acceptance criteria

- The current behavior is demonstrated with repeatable steps before implementation begins.
- The approved solution has a documented source of truth for runtime configuration and does not treat user-controlled environment variables as an authoritative isolation boundary.
- Tests demonstrate that configured memory limits remain effective in supported child-process and SSH/new-shell scenarios.
- Existing supported workloads either remain compatible or have a documented migration path.
- Relevant code and documentation contributions pass project CI and maintainer review.

## Out of scope

- Claiming a container security boundary beyond what HAMi and the underlying NVIDIA/container stack can enforce.
- Committing to a specific configuration-file design before reproduction, threat-model review, and maintainer approval are complete.

## Related issues

- #1112
- #1090

shellyco-code:
## Description

This PR fixes the GPU memory isolation bypass vulnerability that occurs when child processes are spawned (e.g., via `env -i`) or when a user establishes a new SSH session into a HAMi-managed pod.

**The Problem:**
Currently, HAMi enforces GPU isolation by injecting configuration (like `CGROUP_MEMORY_MAX`, `CUDA_VISIBLE_DEVICES`, etc.) via environment variables into the pod's main process. When an SSH session is initiated or environment variables are purged, `libvgpu.so` fails to read these configurations through `getenv()`, effectively bypassing all limits.

**The Solution:**
This PR introduces a thread-safe `vgpu_getenv` helper function that acts as a drop-in replacement for `getenv()`. 
If standard `getenv()` returns `NULL`, `vgpu_getenv` falls back to parsing `/proc/1/environ` (the environment of the container's initialization process). Because Kubernetes always retains the webhook-injected environment variables in PID 1, `libvgpu.so` can securely recover its configuration limits regardless of the user's local shell environment.

### Changes Made
- Added `vgpu_getenv()` implementation in `src/utils.c`.
- Replaced all usages of `getenv()` across `libvgpu.c`, `utils.c`, `log_utils.c`, `multiprocess_memory_limit.c`, and `shrreg_tool.c`.

## Related Issues
Fixes Project-HAMi/HAMi#2125

## Testing
- [x] Verified code compilation without errors.
- [ ] Verified that deploying a pod and running `env -i /bin/bash` followed by a GPU workload respects the original memory limits.
- [ ] Verified that an SSH session into a pod enforces the GPU core utilization and memory boundaries.

## Checklist
- [x] I have read the [contributing guidelines](https://github.com/Project-HAMi/HAMi-core/blob/master/CONTRIBUTING.md).
- [x] My code follows the project's style and conventions.
- [x] I have tested this solution locally.

## Summary by CodeRabbit

* **New Features**
  * Added support for reading configuration values from the primary process environment when they are unavailable in the current environment.
  * Applied this behavior across CUDA visibility, logging, memory limits, resource policies, task priorities, shared-region settings, and library selection.

* **Bug Fixes**
  * Improved configuration handling in multiprocess and containerized environments by providing consistent environment-variable access.

### DIFF
--- a/src/include/utils.h
+++ b/src/include/utils.h
@@ -5,6 +5,7 @@
 #include <unistd.h>
 
 
+char* vgpu_getenv(const char* name);
 int try_lock_unified_lock();
 int try_unlock_unified_lock();
 
--- a/src/libvgpu.c
+++ b/src/libvgpu.c
@@ -92,7 +92,7 @@ FUNC_ATTR_VISIBLE void* dlsym(void* handle, const char* symbol) {
                 break;
             }
         }
-        char *path_search=getenv("CUDA_REDIRECT");
+        char *path_search=vgpu_getenv("CUDA_REDIRECT");
         if ((path_search!=NULL) && (strlen(path_search)>0)){
             vgpulib = dlopen(path_search,RTLD_LAZY);
         }else{
--- a/src/log_utils.c
+++ b/src/log_utils.c
@@ -1,5 +1,6 @@
 #include <stdio.h>
 #include <stdlib.h>
+#include "include/utils.h"
 
 /*
  * Cached log level, read once from LIBCUDA_LOG_LEVEL by log_utils_init().
@@ -10,7 +11,7 @@ int g_log_level = 2;
 FILE *fp1 = NULL;
 
 void log_utils_init(void) {
-    const char *env = getenv("LIBCUDA_LOG_LEVEL");
+    const char *env = vgpu_getenv("LIBCUDA_LOG_LEVEL");
     if (env != NULL) {
         g_log_level = atoi(env);
     }
--- a/src/multiprocess/multiprocess_memory_limit.c
+++ b/src/multiprocess/multiprocess_memory_limit.c
@@ -20,6 +20,7 @@
 #include "include/process_utils.h"
 #include "include/memory_limit.h"
 #include "multiprocess/multiprocess_memory_limit.h"
+#include "../include/utils.h"
 
 
 #ifndef SEM_WAIT_TIME
@@ -93,7 +94,7 @@ void sig_swap_stub(int signo){
 
 // get device memory from env
 size_t get_limit_from_env(const char* env_name) {
-    char* env_limit = getenv(env_name);
+    char* env_limit = vgpu_getenv(env_name);
     if (env_limit == NULL) {
         // fprintf(stderr, "No %s set in environment\n", env_name);
         return 0;
@@ -1028,7 +1029,7 @@ void child_reinit_flag() {
 
 int set_active_oom_killer() {
     char *oom_killer_env;
-    oom_killer_env = getenv("ACTIVE_OOM_KILLER");
+    oom_killer_env = vgpu_getenv("ACTIVE_OOM_KILLER");
     if (oom_killer_env!=NULL){
         if (strcmp(oom_killer_env,"false") == 0)
             return 0;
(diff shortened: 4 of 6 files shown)


lukedev45:
I've been working through the reproduction milestone for #2125 and built a harness for the child/SSH scenarios, so I wanted to flag one observation about the fallback, and offer to run the harness against this branch.

### `/proc/1/environ` isn't always readable

The fallback assumes PID 1 retains the injected environment. That holds when PID 1 is the
workload, but not when PID 1 rewrites its process title and sshd does exactly that. From
my reproduction, with sshd as the container's PID 1:

``
$ docker exec <container> sh -c 'tr "\0" "\n" < /proc/1/environ'
sshd: /usr/sbin/sshd -D -e [listener] 0 of 10-100 startups
``

sshd overwrites the `argv`/`environ` region to show status in `ps`, so the read returns the
title, not the environment. `ps -ef` in the container confirms PID 1 presenting that string.

This matters because sshd-as-PID-1 is a common shape for the deployments this issue targets #1090 describes that scenario. Other daemons that rewrite their title (nginx, postgres)
would behave the same way.

I don't think this sinks the approach, where PID 1 is the workload it should work fine. But
it seems worth handling explicitly or documenting, because it fails silently: `vgpu_getenv`
returns NULL and enforcement falls back to unlimited rather than erroring.

### Related: where the limit actually comes from

While reproducing I found the environment variable isn't the only input, HAMi-core reads the
limit from the shared cache, and the cache takes precedence over the env var in both
directions. Whichever process initialises the cache first sets the limit for the whole
container. Evidence in my comment on
[#2125](https://github.com/Project-HAMi/HAMi/issues/2125).

So it may be worth checking whether this PR alone closes the ordering case, or whether the
precedence rule needs addressing too.

### Offer

My harness covers the original process, forked children, SSH commands, and SSH login shells,
with both orderings. Happy to build this branch and run it against both PID 1 shapes (sshd
as PID 1, and workload-as-PID-1 with sshd as a child) that would cover the regression-test
gap the automated check flagged. Say the word and I'll post results here.

mesutoezdil:

PR #242, opened 2026-07-30
fix: avoid stale process slots after compaction

### LINKED ISSUE #241: Stale process-slot cache can stall TP=2 vLLM after shared-region compaction
## Summary

  A tensor-parallel vLLM deployment could stop generating tokens while using HAMi-core GPU sharing. The affected processes remained alive, but progress stopped after process-slot cleanup in the shared memory region.

  The issue was first observed while starting and running vLLM with tensor parallelism set to 2. Investigation showed that compaction of dead process slots can leave a stale cached slot valid by PID, even though that slot is no longer
  inside the active `proc_num` range. Later status or memory updates may then target an inactive slot, causing accounting to be lost and status synchronization to stall.

  ## Environment

  - HAMi-core base revision: `52f33fc`
  - GPU: NVIDIA RTX PRO 6000 Blackwell Server Edition (SM 12.0)
  - NVIDIA driver: 580.159.03
  - HAMi-core build CUDA version: 12.9.2
  - Runtime CUDA version: 13.0
  - PyTorch: 2.11.0+cu130
  - vLLM: 0.25.1
  - FlashInfer: 0.6.13
  - Workload: Qwen3.6-27B-NVFP4, tensor parallel size 2
  - Two virtual GPUs with 60 GiB memory each

  ## Steps to reproduce

  1. Run a vLLM workload with `--tensor-parallel-size 2` through HAMi-core GPU sharing.
  2. Ensure that multiple CUDA processes share the same HAMi-core memory region.
  3. Allow one process to exit while other processes remain active.
  4. Trigger cleanup of the dead process slot.
  5. Continue CUDA activity in a process whose slot may have been moved during compaction.

  ## Actual behavior

  The workload can stop making progress: vLLM no longer generates tokens, while the involved processes remain alive. Memory accounting and process-status synchronization can target an inactive shared-memory slot.

  ## Expected behavior

  After compaction, every active process must resolve to its slot within the active process range. CUDA work, memory accounting, and process-status synchronization should continue normally.

  ## Proposed fix

  - Clear the original source slot after moving the last live slot during compaction.
  - Validate a cached process slot against the active `proc_num` range, not only by PID.
  - Re-discover the process slot in the active range when the cached slot is stale.
  - Skip process-status waiting when GPU SM limiting is disabled.

  ## Validation performed

  - Built the modified library successfully.
  - Ran the existing CUDA hook consistency check successfully.
  - Verified the updated library in a TP=2 vLLM deployment.
  - Observed the workload under load with `CUDA_DEVICE_SM_LIMIT=50`, `FLASH_ATTN`, and CUDA Graphs disabled (`cudagraph_mode=NONE`). During the observation window, the service remained ready, produced tokens, and had no worker crash or
  engine failure.

AnatoliyKizyulya:
Fixes #241

  ## Summary

  Fix stale cached process slots after shared-region compaction and avoid
  unnecessary status waiting when SM limiting is disabled.

  ## Changes

  - Clear the source slot after moving the last live slot during compaction.
  - Validate cached slots against the active process-slot range.
  - Fall back to an active-slot lookup when the cache is stale.
  - Skip status waiting when no SM limit is active.

  ## Validation

  - CUDA hook consistency check passed: 211 entries and 217 wrappers.
  - Build succeeded in an isolated CUDA 13.3 environment.
  - Validated on NVIDIA RTX PRO 6000 Blackwell Server Edition (SM 12.0),
    driver 580.159.03.
  - vLLM 0.25.1, TP=2, Qwen3.6-27B-NVFP4 remained stable under load with
    `CUDA_DEVICE_SM_LIMIT=0` and `CUDA_DEVICE_SM_LIMIT=50`.

  AI assistance was used for drafting; the change and validation results were reviewed by the author.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved process-slot validation to prevent stale information from affecting GPU status updates and memory accounting.
  * Refined cleanup of inactive process slots to improve the accuracy and reliability of resource tracking.
  * Updated synchronization behavior so waits and retries occur only when GPU core limits are enabled, reducing unnecessary blocking when limits are not in use.
  * Improved handling of per-device GPU limits for more consistent resource management.

### DIFF
--- a/src/include/memory_limit.h
+++ b/src/include/memory_limit.h
@@ -21,7 +21,9 @@ extern int wait_status_self(int status);
 #define ENSURE_RUNNING() {                                \
    /* LOG_DEBUG("Memory op at %d",__LINE__); */              \
     ensure_initialized();                                 \
-    while(!wait_status_self(1)) { LOG_DEBUG("E1"); sleep(1); }             \
+    if (is_gpu_core_limit_enabled()) {                    \
+        while(!wait_status_self(1)) { LOG_DEBUG("E1"); sleep(1); }         \
+    }                                                      \
 }                                                         \
 
 #define INC_MEMORY_OR_RETURN_ERROR(bytes) {               \
--- a/src/multiprocess/multiprocess_memory_limit.c
+++ b/src/multiprocess/multiprocess_memory_limit.c
@@ -7,6 +7,7 @@
 #include <stdlib.h>
 #include <errno.h>
 #include <stddef.h>
+#include <stdint.h>
 #include <semaphore.h>
 #include <unistd.h>
 #include <time.h>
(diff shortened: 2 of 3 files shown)


AnatoliyKizyulya:
@coderabbitai review

AnatoliyKizyulya:
This patch also resolves the issue of running models on vLLM 0.27.1 with TP=2. It was tested on an H200 GPU using driver version 580.178.04.

AnatoliyKizyulya:
@archlitchi could you please review this PR?

ji-cy:
I hit the same issue and used an LLM to fix it. Here are the changes — self-tested and working. FYI.

https://github.com/ji-cy/HAMi-core/commit/af13dfc9b1d4e41b057bf3712ef818d509c73c9a

Also: why need to move the slot?

AnatoliyKizyulya:
> I hit the same issue and used an LLM to fix it. Here are the changes — self-tested and working. FYI.
> 
> [ji-cy@af13dfc](https://github.com/ji-cy/HAMi-core/commit/af13dfc9b1d4e41b057bf3712ef818d509c73c9a)
> 
> Also: why need to move the slot?

Good question: moving the slot is not fundamentally required, and this PR did not introduce that behavior. clear_proc_slot_nolock() already used swap-with-last compaction.
The reason is the existing representation invariant: after cleanup, live processes are expected to occupy the dense prefix procs[0..proc_num), and a new process is appended at procs[proc_num]. If slot i is removed and it is not the
last slot, simply decrementing proc_num would leave the last live process outside the active range while keeping the hole inside it. Moving the last live slot into i preserves the dense prefix and provides O(1) removal. Once it is
moved, the old tail slot must also be cleared; otherwise it retains a duplicate PID and stale accounting data.
Your in-place hole-reuse approach is a valid alternative and has an important advantage: live slots keep stable addresses, so cached my_slot pointers are not invalidated. However, it changes proc_num from the end of a packed live range
into a high-water mark containing possible pid == 0 holes. That requires every iterator, accounting path, status check, and signal loop to handle holes, registration to reuse them, and calls such as kill() to explicitly avoid PID 0.
This PR intentionally preserved the existing packed-array semantics and fixed stale-cache handling around the pre-existing compaction. I agree that stable, non-moving slots may be a cleaner long-term design, but it is a broader
behavioral change and needs a complete audit and concurrency testing.

archlitchi (approved):
/lgtm

PR #249, opened 2026-08-04
chore: remove unused set_current_device_sm_limit setters

### LINKED ISSUE #231: chore: remove the sm scale
https://github.com/Project-HAMi/HAMi-core/blob/52f33fc7fa1fbb3f08148ab076d0e7447bec7f2a/src/multiprocess/multiprocess_memory_limit.h#L132

I am checking the hami-core code and this `scale` is unclear to me.

It is not used anywhere and I do not understand why we need a `scale` to set sm limit.

Should this be removed?

veyron-kairo:
## What this does

Removes two unused functions flagged in #231:

- `set_current_device_sm_limit(int dev, int scale)` — declared in the header but never defined or called.
- `set_current_device_sm_limit_scale(int dev, int scale)` — defined but never called anywhere in the tree, including `test/`.

A repo-wide search confirms neither symbol is referenced outside its own declaration/definition.

The `sm_init_flag` field that the scale setter used to touch is intentionally left in the shared-region struct — it is part of the memory-mapped, cross-process layout, so removing it would be a compatibility change beyond the scope of this cleanup. With the setter gone it is only initialized to `0` and never read, which is inert.

Credit to @KaminariOS for spotting these in #231.

## Verification

Repo-wide `grep` confirms no remaining references. I was not able to run a full build locally (it needs the CUDA toolkit), but since the removed symbols have no callers or other definitions, the removal cannot affect compilation — the `build-in-docker` target / CI will confirm.

Fixes #231

## Summary by CodeRabbit

* **Refactor**
  * Removed legacy controls for adjusting device SM memory limits.
  * No end-user visible behavior changes are expected.

### DIFF
--- a/src/multiprocess/multiprocess_memory_limit.c
+++ b/src/multiprocess/multiprocess_memory_limit.c
@@ -1386,18 +1386,6 @@ int set_host_pid(int hostpid) {
     return 0;
 }
 
-int set_current_device_sm_limit_scale(int dev, int scale) {
-    ensure_initialized();
-    if (region_info.shared_region->sm_init_flag==1) return 0;
-    if (dev < 0 || dev >= CUDA_DEVICE_MAX_COUNT) {
-        LOG_ERROR("Illegal device id: %d", dev);
-    }
-    LOG_INFO("dev %d new sm limit set mul by %d",dev,scale);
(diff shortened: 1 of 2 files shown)


mesutoezdil:

PR #251, opened 2026-08-06
feat: add host PID broker client and CUDA context accounting

iemAnshuman:
Related issue: Project-HAMi/HAMi#1662

Design discussion: Project-HAMi/HAMi#2244

Companion server PR: Project-HAMi/HAMi#2417

Current companion server head: `146ee027b8b0ac44c8f8b3d48581f2cf69a485e9`

## What this changes

This adds an optional host PID broker client to HAMi-core. When `LIBVGPU_HOSTPID_BROKER=1`, `postInit()` asks the HAMi device plugin for the caller's host PID over a Unix socket. A valid response avoids the temporary CUDA primary context probe used by the NVML discovery path.

The client accepts only `/tmp/vgpulock/hostpid/broker.sock`. It verifies the broker directory, read only mount, socket owner, server UID, protocol fields, deadline, and returned PID before accepting a response.

The feature is disabled by default. A new HAMi-core build without the companion server continues through the NVML fallback.

## Fallback correctness

The fallback must serialize NVML discovery across independent cache files because the NVML snapshots are node wide. The correction takes one node wide lock on the trusted broker mount before the existing cache record lock. It never accepts a guessed PID.

If the broker gate is enabled but the trusted mount is missing or unsafe, host PID discovery fails clearly. It does not return to cache local discovery, which could accept a concurrently starting peer PID.

The final path hardening also checks every path component, the supported filesystem policy, permission changes while a waiter is blocked, object replacement, and one absolute deadline shared by the node and cache locks.

## CUDA context accounting

Removing the temporary probe also removes its context size measurement. The broker path measures memory when the application first retains the real primary context.

Accounting maps visible CUDA ordinals to physical NVML devices by PCI identity. It keeps retain and charge state for each visible device, clears inherited state after `fork()`, charges nested retains once, and removes the charge after the final successful release.

## Validation

1. Current client head `4421b0970c172e343c7fb59491952b78ea9d86eb` passes the GitHub library build, hook consistency, cpplint, DCO, CodeRabbit, dependency, license, and security checks.

2. Exact public heads `8d59bf9` and `935a6de` completed all 16 balanced eight GPU `cuInit()` cells in job `181726`. Mean API wall time changed from 56.373 to 22.501 seconds at N=128 and from 130.662 to 50.288 seconds at N=300. Every worker completed. This measures `cuInit()` only.

3. Jobs `181845` and `182609` show the remaining driver cost. Without admission, the broker condition took 155.152 seconds at N=128 and 792.573 seconds at N=240. The N=300 cell reached its 903.11 second deadline. All 300 broker queries completed with zero broker failures, so the timeout was after PID discovery.

4. Fallback job `182607` tested revision `b0df70a` with one and eight GPU layouts at N=128 and with N=300. Every candidate cell recorded zero peer and zero external PID mappings. The candidate failed discovery clearly for 71 of 128, 116 of 128, and 290 of 300 workers in those three cells. These are correctness results, not completion or performance claims.

5. The fallback lock tests cover owner death, live holder timeout, independent caches, `fork()`, `exec()`, and overlap. A 300 waiter owner death run completed with no premature or failed waiter.

6. The Linux root path hardening run completed 28 recorded checks with zero failures. Its archive SHA256 is `941d0a7c3de6e4602a8c57f3e71b8c21ad3acc60cef127d07599ef8ed6f7783a`. The archive covers the hardening source through `607d7e83c282f8c494b39a0c2d92e16a83bdf965`. Later review and lint commits at the current public head are outside that archive.

7. The source and evidence package passes its checksum, semantic, archive, and adversarial verifier suites. It reports 46 verified artifacts, 0 pending artifacts, and 0 failures. The package ledger SHA256 is `294637ae0abbe69115c9f900442029fa0ccb335ec08ad3e8982d17f8dea1448a`.

The [interim validation report](https://github.com/Project-HAMi/HAMi/pull/2417#issuecomment-5256411883) records the source boundaries, A100 results, fallback result, Linux archive, and remaining Kubernetes checks.

## Remaining work

1. Real kubelet allocation with the broker enabled and disabled, DaemonSet rollout and process restart, rollback, and degradation when the socket is unavailable remain untested on an isolated NVIDIA Kubernetes node.

2. Additional container runtimes and other GPU and driver versions remain untested. The A100 result above predates the later fallback hardening commits at the current public head.

3. The broker removes PID discovery from the serialized path, but it does not remove primary context contention. Driver admission is an optional separate follow up and remains disabled by default.

## User facing change

This adds the `LIBVGPU_HOSTPID_BROKER=1` opt in. Default behavior remains unchanged.

## AI assistance

I used some AI assistance for this.


mesutoezdil:
resolve conflicts

iemAnshuman:
> resolve conflicts

conflicts are resolved. If this PR is too large for review, I can split it into smaller prs.

maverick-woo:
Thanks for the work. Please split this into smaller, focused PRs. It would also be best to provide a complete HAMi end-to-end test covering the real workload path, not only `cuInit()`.

iemAnshuman:
> Thanks for the work. Please split this into smaller, focused PRs. It would also be best to provide a complete HAMi end-to-end test covering the real workload path, not only `cuInit()`.

makes sense. converting to draft and splitting into context accounting, fallback lock, broker client, then the integration.

mesutoezdil:
bcs of the inactivity was closed

Issue #252, opened 2026-08-07
Shared-region init serialises across concurrent processes: p50 grows 0.25ms to 5.6ms from 1 to 64 processes


keshav9926:
I've been reading the shared-region code and wanted to put numbers on how first-touch initialisation behaves when many processes start at once, since that's the shape a TP=N inference job or a multi-container pod produces. Sharing the measurements and what I think the mechanism is, in case it's useful.

## Measurements

Each process is held at a shared start barrier, released together, then times exactly one `ensure_initialized()` call. The shared region is deleted before every round, so each round measures cold first-touch. 5 repeats per point.

| procs | wall_ms | init_p50 | init_p95 | init_max |
|------:|--------:|---------:|---------:|---------:|
| 1 | 0.78 | 0.25 | 0.67 | 0.67 |
| 2 | 0.92 | 0.40 | 0.60 | 0.60 |
| 4 | 1.24 | 0.60 | 1.23 | 1.23 |
| 8 | 5.51 | 0.82 | 10.45 | 10.56 |
| 16 | 6.97 | 1.89 | 9.57 | 9.98 |
| 32 | 10.28 | 2.59 | 5.88 | 7.13 |
| 64 | 18.26 | 5.62 | 18.69 | 34.27 |

If initialisation were concurrent, `init_p50` would stay roughly flat as `procs` grows. It doesn't â€” p50 grows ~22x and the tail (`init_max`) reaches 34ms. The knee is between 4 and 8 processes. A second independent run reproduced the same shape (32 procs: p50 3.44ms vs 2.59ms here).

## Mechanism, as far as I can tell

Two serialisation points on the init path:

**1. `try_create_shrreg()` holds a whole-file advisory lock across region setup.**
`lockf(fd, F_LOCK, SHARED_REGION_SIZE_MAGIC)` at `multiprocess_memory_limit.c:1266` is held until `F_ULOCK` at :1327. Every process doing first-touch queues here.

**2. `init_proc_slot_withlock()` does O(N) work under the global semaphore.**
It takes `lock_shrreg()` (a single `sem_timedwait` over the whole region), and inside, `clear_proc_slot_nolock()` walks every occupied slot calling `proc_alive()` â€” which is an `fopen("/proc/<pid>/stat")` + read + close per slot (`src/include/process_utils.h:17`).

[...]

PR #253, opened 2026-08-08
fix: sweep dead process slots only when slots are scarce

keshav9926:
Fixes the O(N²) join path measured in #252.

## What the problem is

`init_proc_slot_withlock()` calls `clear_proc_slot_nolock(1)` on every join. That sweep walks every occupied slot and calls `proc_alive()`, which is an `fopen("/proc/<pid>/stat")` + read + close per slot (`src/include/process_utils.h:17`). All of it runs under `lock_shrreg()`, the single region-wide semaphore.

So the Nth process to join does O(N) filesystem syscalls while holding a lock every other joining process needs. Across N processes starting together that is O(N²) serialised work — the shape a TP=N inference job or a pod whose containers all start at once produces.

## What this changes

A join now sweeps for liveness only once occupancy reaches three quarters of the table.

Why that is safe: reclaiming a slot whose process died is not needed for a join to be correct. It matters in two places, and both are still covered.

- **Memory accounting.** A dead process's slot inflates `get_gpu_memory_usage()`, which could cause a spurious OOM. `oom_check()` already calls `clear_proc_slot_nolock(1)` and retries before it reports OOM (`src/allocator/allocator.c:54`), so the reclaim still happens exactly where a stale slot changes an outcome.
- **Slot exhaustion.** Handled by the threshold. The sweep now runs *before* the `proc_num >= SHARED_REGION_MAX_PROCESS_NUM` capacity check instead of after the insert, so a table filled with dead slots is recovered rather than hitting `exit_withlock(-1)`. That case used to be fatal.

Slots that exit cleanup already marked with PID 0 are still compacted on every join. That path reads no files, so it costs nothing.

## Measurements

Harness from #252 — fork N children, hold them at a shared start barrier, release them together, time exactly one `ensure_initialized()` per child; the region is deleted before every round so each round measures cold first-touch: https://gist.github.com/keshav9926/9e8f4c29eebd104ba02891e4733b9dfb

Both arms were built from this tree, one with the patch and one without, and run **interleaved** — base, fix, base, fix — for 6 iterations of 10 repeats each, so thermal drift and background load hit both arms equally. Values below are the median across the 6 iterations.

| procs | init p50 before | init p50 after | |
|------:|----------------:|---------------:|-----:|
| 1  | 0.35 ms | 0.27 ms | 1.3× |
| 2  | 0.43 ms | 0.30 ms | 1.4× |
| 4  | 0.64 ms | 0.42 ms | 1.5× |
| 8  | 1.03 ms | 0.63 ms | 1.6× |
| 16 | 1.42 ms | 0.71 ms | 2.0× |
| 32 | 2.66 ms | 0.99 ms | 2.7× |
| 64 | 6.06 ms | 1.11 ms | 5.5× |

Before, p50 grows 17× from 1 to 64 processes. After, it grows 4×, and the gap widens with process count — which is what you would expect if the removed term was the quadratic one. The remaining growth is the `lockf` in `try_create_shrreg()` and the semaphore itself; this PR does not touch either.

`wall_ms`, `init_p95` and `init_max` are dominated by fork/exit scheduling noise on this hardware and I would not read anything into them.

## Test

`test/test_proc_slot_reclaim.c` — GPU-free, built against the production shared-region sources the same way `test_postinit_owner_death` is, and registered with ctest.

It pins both halves of the new contract:

- below the threshold, a join leaves slots held by dead processes in place (children are SIGKILL'd, so exit cleanup never runs and the slot keeps a PID that no longer exists);
- at the threshold, a join reclaims them, so repeated join-and-die cycles cannot grow the table without bound.

The target is compiled with `SHARED_REGION_SWEEP_THRESHOLD=8` so the reclaim path is reachable without spawning 768 processes. The production constant is unchanged; it is now `#ifndef`-guarded only so the test can override it.

Verified red/green: against `main` without the fix the test fails (`join below the sweep threshold changed the table: expected 4 occupied slots, saw 2`); with the fix it passes 5/5 consecutive runs.

## What was tested, and what wasn't

- Environment: WSL2 (Ubuntu 24.04, kernel 6.6.87), NVIDIA GeForce RTX 3050 Laptop GPU, driver 592.82, single GPU, driver libs via `/usr/lib/wsl/lib`.
- This change does not touch device allocation or in-container isolation — it only changes *when* the slot table is swept — so I did not run the GPU allocation suite against it. I have not run it on a multi-GPU datacenter node, and `/proc` and file-lock behaviour under WSL2 may differ from bare metal. Happy to re-run anywhere if someone has a node available.
- Only `ensure_initialized()` is timed. No CUDA context creation, no real workload.
- The absolute numbers are small. What I think matters is the shape: the cost scaled with process count, which is the thing operators scale up.

## Trade-off worth flagging

Between sweeps, slots held by dead processes stay in the table, so `proc_num` and the usage totals derived from it can be stale for longer than before. `oom_check()` covers the case where that changes an allocation outcome. If you would rather also bound the staleness in wall-clock terms, a "sweep if the last one was more than N seconds ago" rule would do it, but that needs a timestamp in the shared region and therefore a layout change — happy to do it that way instead if you prefer.

(Disclosure: I use AI assistance in my workflow. The measurements, the code reading and the reasoning above are my own, and I'm happy to walk through any part of it.)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recovery of shared resources after worker processes terminate unexpectedly.
  - Reduced unnecessary process checks during normal worker cleanup, improving reliability when joining shared regions.
  - Added threshold-based reclamation to prevent exhausted process slots from blocking new workers.

- **Tests**
  - Added regression coverage for reclaiming slots from forcibly terminated workers, including timeout and cleanup handling.

### DIFF
--- a/src/multiprocess/multiprocess_memory_limit.c
+++ b/src/multiprocess/multiprocess_memory_limit.c
@@ -1072,6 +1072,20 @@ void init_proc_slot_withlock() {
     shared_region_t* region = region_info.shared_region;
 
     int proc_num = atomic_load_explicit(&region->proc_num, memory_order_acquire);
+
+    // A full sweep reads /proc/<pid>/stat once per occupied slot, so it costs
+    // O(proc_num) filesystem syscalls while the region lock is held. Running it
+    // on every join makes N processes starting together O(N^2) serialised work.
+    // Reclaiming a slot whose process already died is not needed for the join
+    // itself to be correct: oom_check() sweeps before it reports OOM, which is
+    // where a stale slot actually changes an outcome. So sweep here only when
+    // slots are scarce -- and do it before deciding the table is full, so a
+    // table filled with dead slots is recovered instead of being fatal.
+    if (proc_num >= SHARED_REGION_SWEEP_THRESHOLD) {
+        clear_proc_slot_nolock(1);
+        proc_num = atomic_load_explicit(&region->proc_num, memory_order_acquire);
+    }
+
     if (proc_num >= SHARED_REGION_MAX_PROCESS_NUM) {
         exit_withlock(-1);
     }
@@ -1123,7 +1137,9 @@ void init_proc_slot_withlock() {
         atomic_fetch_add_explicit(&region->proc_num, 1, memory_order_release);
     }
 
-    clear_proc_slot_nolock(1);
+    // Slots that exit cleanup already marked dead carry PID 0, and dropping
+    // those reads no files at all, so that part stays on the join path.
+    clear_proc_slot_nolock(0);
     unlock_shrreg();
 }
 
--- a/src/multiprocess/multiprocess_memory_limit.h
+++ b/src/multiprocess/multiprocess_memory_limit.h
@@ -40,6 +40,12 @@
 
 #define SHARED_REGION_SIZE_MAGIC  sizeof(shared_region_t)
 #define SHARED_REGION_MAX_PROCESS_NUM 1024
+// Slot-table occupancy at which joining a process performs a full liveness
+// sweep. See init_proc_slot_withlock(). Overridable at build time so the
+// regression test can reach the sweep without spawning 768 processes.
+#ifndef SHARED_REGION_SWEEP_THRESHOLD
+    #define SHARED_REGION_SWEEP_THRESHOLD ((SHARED_REGION_MAX_PROCESS_NUM * 3) / 4)
+#endif
 
 // macros for debugging
 #define SEQ_FIX_SHRREG_ACQUIRE_FLOCK_OK 0
--- a/src/multiprocess/multiprocess_memory_limit.h
+++ b/src/multiprocess/multiprocess_memory_limit.h
@@ -46,6 +46,12 @@
 #ifndef SHARED_REGION_SWEEP_THRESHOLD
     #define SHARED_REGION_SWEEP_THRESHOLD ((SHARED_REGION_MAX_PROCESS_NUM * 3) / 4)
 #endif
+// Past capacity the sweep could never run, so a table full of dead slots would
+// reach the capacity check and exit. At zero every join sweeps again.
+#if SHARED_REGION_SWEEP_THRESHOLD < 1 || \
+    SHARED_REGION_SWEEP_THRESHOLD > SHARED_REGION_MAX_PROCESS_NUM
+    #error "SHARED_REGION_SWEEP_THRESHOLD must be between 1 and SHARED_REGION_MAX_PROCESS_NUM"
+#endif
 
 // macros for debugging
 #define SEQ_FIX_SHRREG_ACQUIRE_FLOCK_OK 0
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -14,20 +14,31 @@ foreach(TEST_SCRIPT ${TEST_SCRIPTS})
     get_filename_component(TEST_TARGET_NAME ${RELATIVE_TEST_PATH} NAME_WE)
     set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${TEST_TARGET_DIR})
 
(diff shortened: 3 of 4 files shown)


mesutoezdil:
resolve conflicts pls

maverick-woo:
Could dead slots below the threshold still be counted by get_gpu_memory_usage(), causing cuMemGetInfo/NVML to report stale usage before oom_check() can clean them?

badnikhil:
@keshav9926 I ran your change against a wider N because I wanted the same curve on a machine
without a GPU, and it holds well past the range in #252. Sharing the numbers in case they are
useful for the rebase.

Harness: N processes fork+exec'd, each dlopens libvgpu.so and resolves `try_create_shrreg` and
`init_proc_slot_withlock`, then all block on a shared pipe and are released together. No CUDA and
no NVML are involved — libvgpu.so has no DT_NEEDED on libcuda or libnvidia-ml and neither of those
two functions calls into the driver, so this runs on a laptop. `lockf` and `sem_timedwait` are
interposed from the harness executable, so time spent waiting for a lock is separated from time
spent holding it. Workers stay alive until the parent has collected every result, otherwise they
become zombies, `proc_alive` evicts their slots, and the sweep being measured disappears. Cache
file on tmpfs, fresh per point, three runs each, medians below.

On main (f01e9f2), the time spent holding the region semaphore inside `init_proc_slot_withlock`:

    N       1     8    32    64   128   256   384   512
    crit  0.06  0.11  0.44  0.86  0.90  3.01  5.08  5.09   (ms, p50)

Dividing by N gives 12 to 14 microseconds per occupied slot, flat from N=8 to N=512. That is one
`fopen`/`fscanf`/`fclose` of `/proc/<pid>/stat`, which is what `proc_alive` does, so it is a
syscall count rather than a queueing effect and should not depend on core count. It puts a floor of
roughly `12us * N` on every joiner and `12us * N^2` on a simultaneous start, independent of
hardware: about 1.1s at the 300 processes in HAMi#1662, about 12s at the 1024-slot ceiling.

With your change (threshold 768, `clear_proc_slot_nolock(0)` on the join path):

    N            64     128     256     384     512
    main p99  120.30  104.61  780.80 1808.08 2298.69   (ms)
    #253 p99    0.86   10.48   11.96  314.82   74.10
    main crit  0.856   0.899   3.007   5.084   5.091
    #253 crit  0.009   0.012   0.024   0.029   0.034

The critical section stops scaling — 5.09ms to 0.034ms at N=512, and flat in N. Wall clock for the
whole shared-region path at N=512 goes from 2.77s to 0.20s.

Two things worth flagging.

First, your point (1) is confirmed, but only after (2) is fixed. On main the `lockf(F_LOCK)` in
`try_create_shrreg` never exceeds 25ms p99 anywhere in the sweep, so it is invisible behind the
semaphore. Once your change removes the sweep it becomes the entire residual tail: p99 314ms at
N=384 and 47ms at N=512, with the critical section at 30 microseconds. It is also visibly
non-monotonic across repetitions, so I would not build anything on it yet.

Second, on @maverick-woo's question about stale usage below the threshold, I think the answer is
yes but only for reported numbers, not for decisions. `get_gpu_memory_usage()` sums
`slot->used[dev].total` over `i < proc_num` with no liveness check at all; it reads `pid` and
`hostpid` only to log them. A SIGKILL'd process never runs `exit_handler`, so its slot keeps a
non-zero pid and a non-zero charge. Under this PR nothing reclaims it below 768 slots except
`oom_check()`, which does take the lock, sweep, and re-check before declaring OOM, and recurses
after a successful clear. So allocation decisions still see a swept table at the moment it matters,
but the lock-free readers behind `cuMemGetInfo` and the NVML hooks will over-report until an OOM
check happens or the table reaches the threshold.

Caveats on my side: 8 laptop cores, so N=512 is 64:1 oversubscription and the queueing on top of
the O(N) term is amplified relative to a real node; and this measures only the shared-region path,
not `set_task_pid`, which needs a real driver. Happy to rerun any point or share the harness.

mesutoezdil:

PR #258, opened 2026-08-09
fix: forward untracked pointers to the driver on the synchronous free path

### LINKED ISSUE #257: cuMemFree of an untracked pointer leaks the allocation and returns an invalid CUresult
## What happens

`remove_chunk` in `src/allocator/allocator.c` is the synchronous free path behind `cuMemFree_v2` -> `free_raw`. It returns `-1` without freeing when the pointer isn't found in `device_overallocated`, in two places:

- the empty-list early return: `if (a_list->length == 0) return -1;`
- the loop not-found return at the end of the function

`free_raw` and the `cuMemFree_v2` hook pass that `-1` straight back to the caller. So when an app frees a device pointer that libvgpu didn't track in that list, two things go wrong:

1. the allocation is never freed on the device, so it leaks, and
2. the caller gets `-1`, which isn't a valid `CUresult` and shows up as an "unrecognized error code".

## How a pointer reaches cuMemFree untracked

- a stream-ordered allocation (`cuMemAllocAsync`, which is tracked in the separate `device_allocasync` list) that the app frees with `cuMemFree` instead of `cuMemFreeAsync`
- an allocation made before libvgpu started tracking

## The async twin was already fixed

`remove_chunk_async` had the same bug and was fixed to forward the free to the real driver on the not-found path. Its comment says:

> Not tracked by libvgpu: free it for real instead of returning -1, which leaked and surfaced as an "unrecognized error code".

The synchronous `remove_chunk` never got that same fix.

## A related accounting concern

While tracing this I noticed a deeper inconsistency that's probably worth a maintainer's view. Sync allocations (`device_overallocated`, charged the exact size) and async allocations (`device_allocasync`, charged against a pool high-water mark) live in two separate lists, and each free path only searches its own list. When a pointer crosses paths, the memory does get freed but `rm_gpu_device_memory_usage` is never called for it, so the per-process accounting that enforces the vGPU memory limit drifts upward over time and can eventually cause a false OOM below the real limit.

The PR only fixes the clear leak and invalid return on the sync path (mirroring the async fix). I'm happy to look into the accounting drift separately if that's useful.

veyron-kairo:
## What this does

`remove_chunk` is the synchronous free path behind `cuMemFree_v2` -> `free_raw`. It returned `-1` without freeing when a pointer wasn't found in `device_overallocated`, both on the empty-list early return and on the loop not-found return. That `-1` propagates back through `free_raw` and the `cuMemFree_v2` hook, so freeing a pointer libvgpu didn't track leaves the memory live on the device and hands the caller a value that isn't a valid `CUresult`.

`remove_chunk_async` already forwards such frees to the real driver; its comment notes the old behavior "leaked and surfaced as an unrecognized error code". This applies the same fix to the synchronous path, using the existing `cuMemoryFree` helper that the found path already calls.

Fixes #257

## How a pointer reaches this path untracked

- a stream-ordered allocation (`cuMemAllocAsync`) freed with `cuMemFree` instead of `cuMemFreeAsync`
- an allocation made before tracking started

## Verification

The change reuses `cuMemoryFree(dptr)`, which `remove_chunk` already calls on its found path, so it adds no new dependency and is build-identical to existing code. I couldn't run a full build locally (it needs the CUDA toolkit); the Build libvgpu CI will confirm compilation.

I used AI assistance while reviewing the allocator . I read the code and the async twin's fix myself and take responsibility for the change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory cleanup for untracked device-pointer allocations.
  * Device memory can now be freed correctly even when no matching allocation record exists.

### DIFF
--- a/src/allocator/allocator.c
+++ b/src/allocator/allocator.c
@@ -181,7 +181,10 @@ int remove_chunk(allocated_list *a_list, CUdeviceptr dptr) {
     CUdevice t_dev;
 
     if (a_list->length == 0) {
-        return -1;
+        /* Nothing tracked here, so the pointer was not allocated through this
+         * list. Forward the free to the real driver rather than returning -1
+         * (see the not-found path below). */
+        return cuMemoryFree(dptr);
     }
 
     pthread_mutex_lock(&mutex);
@@ -203,7 +206,12 @@ int remove_chunk(allocated_list *a_list, CUdeviceptr dptr) {
     }
 
     pthread_mutex_unlock(&mutex);
-    return -1;
+    /* Not tracked in this list (e.g. a stream-ordered allocation reaching the
+     * synchronous free path): forward the free to the real driver instead of
+     * returning -1, which leaves the allocation live and surfaces to the caller
+     * as an unrecognized CUresult. Mirrors the not-found path in
+     * remove_chunk_async. */
+    return cuMemoryFree(dptr);
 }
 
 int remove_chunk_only(CUdeviceptr dptr) {


src/allocator/allocator.c
```diff
@@ -181,7 +181,10 @@ int remove_chunk(allocated_list *a_list, CUdeviceptr dptr) {
     CUdevice t_dev;
 
     if (a_list->length == 0) {
-        return -1;
+        /* Nothing tracked here, so the pointer was not allocated through this
```
archlitchi, line 184: please remove these comments and comments below

maverick-woo:
Looks good overall. Please move the empty-list check under the mutex and remove the verbose comments.

archlitchi (approved):
/lgtm

Issue #262, opened 2026-08-10
Build fails on current toolchains (GCC 15/C23 and CMake 4)


Smallfu666:
## Problem

HAMi-core `main` at `5496322` does not build with a current native toolchain.

Environment: Ubuntu 26.04, GCC 15.2.0 (default `gnu23`), CMake 4.2.3, and CUDA toolkit 13.3.

- CMake 4 rejects the top-level and `src/` minimum versions (2.8.12 and 2.8.11) because compatibility with CMake versions below 3.5 has been removed.
- In C23, an empty function parameter list `()` is a zero-argument prototype. The common CUDA/NVML hook pointer typedefs therefore make calls with real API arguments fail; together with the incompatible `pthread_create` callback declaration, a full build reports 452 errors.
- `test/CMakeLists.txt` still uses the deprecated FindCUDA module and `cuda_add_executable`. FindCUDA is unavailable when CMP0146 is set to `NEW`; with the project's current policy state on CMake 4.2, it emits a CMP0146 warning and the legacy discovery path fails here unless `CUDA_TOOLKIT_ROOT_DIR` is supplied.

## Proposed fix

- Dispatch each hook through the symbol's real type using `typeof(&sym)`, add the required forward declarations, and give the watcher callback its required `void *arg` parameter.
- Raise both CMake minimum versions to 3.18 for CMP0104 and `CMAKE_CUDA_ARCHITECTURES` behavior. The current UBI8 build image provides CMake 3.26.5.
- Replace FindCUDA with `enable_language(CUDA)`, FindCUDAToolkit, and the `CUDA::cuda_driver` / `CUDA::cudart` imported targets while preserving shared `cudart` linkage.

## Validation

A clean build of the prepared fix with the toolchain above completes with exit 0 and no compiler errors, produces `libvgpu.so`, and builds all 19 test targets. The fix is available in PR #263.

PR #268, opened 2026-08-11
fix: use CUDA_DEVICE_MAX_COUNT for ctx_activate declaration

AshrafAhmed9:
`ctx_activate` is defined as `int[32]` in `multiprocess_memory_limit.c` but declared `extern int[16]` in `cuda/context.c`. Incompatible declarations of the same object are undefined behaviour, and it's the kind of mismatch LTO builds and newer compilers diagnose — relevant given #262.

Every other per-device array in the tree is sized with `CUDA_DEVICE_MAX_COUNT` and the device loops bound on it, so this uses the constant in both places instead of either magic number. `context.c` already includes the header that defines it.

No behaviour change: nothing indexes `ctx_activate` above `CUDA_DEVICE_MAX_COUNT`.

### DIFF
--- a/src/cuda/context.c
+++ b/src/cuda/context.c
@@ -2,7 +2,7 @@
 #include "multiprocess/multiprocess_memory_limit.h"
 
 extern size_t context_size;
-extern int ctx_activate[16];
+extern int ctx_activate[CUDA_DEVICE_MAX_COUNT];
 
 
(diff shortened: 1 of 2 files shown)


mesutoezdil:

AshrafAhmed9:
Understood, no problem. Could you point me at which guideline this fell under so I get it right next time?

PR #272, opened 2026-08-12
test: GPU-free regression test for env/SSH memory-limit isolation (#2125)

mohitt31:
## Description

Adds a hardware-free CI characterization test for the memory-limit isolation gap in [Project-HAMi/HAMi#2125](https://github.com/Project-HAMi/HAMi/issues/2125).

`test/test_env_isolation.c` links the production shared-region implementation directly and calls **no** CUDA/NVML entry point, so it runs in CI without a GPU — the same GPU-free linking pattern as the existing `test/test_postinit_owner_death.c` (`--gc-sections` drops the GPU-facing paths).

It pins `get_current_device_memory_limit(0)` — the value `oom_check()` gates every `cudaMalloc` on — across the process-startup shapes from the issue:

| Case | Startup shape | Enforced limit |
|---|---|---|
| [A] | configured process, fresh region | 1 GiB (enforced) |
| [C] | env-stripped process, first writer of a fresh region | 0 (leak) |
| [D] | configured workload joining a region an env-less process seeded with 0 first | 0 (leak) |

## Notes

- Assertions encode **current** behavior, so the test is green on today's code and documents the boundary.
- The two leak assertions are tagged `[FIX FLIPS THIS]`; whichever fix maintainers choose flips them from `0` to the configured limit, turning this into the issue's acceptance test.
- Intentionally **fix-agnostic** — a CI test for whoever implements the fix, not a fix itself.

Repro harness (also GPU-free): https://gist.github.com/mohitt31/a6a825f429ccce3edbaafe50425fa37c

## Testing

Built and run locally against the production shared-region source; all four assertions pass (`env-isolation characterization tests passed`, exit 0).

## Summary by CodeRabbit

* **New Features**
  * Added coverage for shared-memory limit behavior across isolated processes.
  * Added scenarios for configured limits, environment-stripped processes, and later processes affected by earlier configuration.
  * Added bounded execution, resource cleanup, and failure reporting for these checks.

* **Tests**
  * Added a GPU-free automated test with a 30-second timeout.
  * Updated test build settings to use a focused, non-CUDA configuration.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -12,7 +12,8 @@ foreach(TEST_SCRIPT ${TEST_SCRIPTS})
     get_filename_component(TEST_TARGET_NAME ${RELATIVE_TEST_PATH} NAME_WE)
     set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${TEST_TARGET_DIR})
 
-    if (TEST_TARGET_NAME STREQUAL "test_postinit_owner_death")
+    if (TEST_TARGET_NAME STREQUAL "test_postinit_owner_death" OR
+        TEST_TARGET_NAME STREQUAL "test_env_isolation")
         # Build this focused regression test with the production shared-region
         # implementation.  It does not invoke any CUDA/NVML entry point at
         # runtime.  Section garbage collection drops unrelated GPU-facing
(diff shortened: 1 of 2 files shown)


mohitt31:
Thanks for the review. Addressed both nitpicks in the latest push:

- `test/test_env_isolation.c`: `wait_child_bounded` now `SIGKILL`s and reaps the child on timeout before returning, so a stuck probe can't leak shared-region/cache state into a later probe (added `<signal.h>`).
- `test/CMakeLists.txt`: raised the `env_isolation` CTest timeout from 20s to 30s, above the test's own 4×5s worst-case budget, so a slow run reports the mismatch instead of being killed.

Rebuilt and re-ran locally — all four assertions still pass.

mesutoezdil:

PR #279, opened 2026-08-15
fix repeated RTLD_NEXT lookups returning null

iemAnshuman:
## Why

repeated calls to dlsym(RTLD_NEXT, "open") return the same libc function. libvgpu.so is preloaded and the first lookup works but the second one returns null.

The wrapper keeps completed (thread, pointer) pairs in a global table. check_dlmap() considers it recursion if a later lookup returns the same pointer on the same thread. The table tracks completed calls so it cannot tell the difference between recursive entry and a normal repeated lookup.

This patch removes that table and it mutex. Now the path thru RTLD_NEXT returns the result from the resolved libc dlsym directly. It also adds a regression test for the failing case, both normal and with libvgpu.so preloaded.

 ## Testing

  1. main current reproduced second lookup failure in 5 of 5 runs. The patched library passes 5 out of 5 runs.
  2. The patched library completed 40,000 lookups on four symbols with eight threads
  3. Both focused CTests ran 20 times each successfully.
  4. A direct cuda driver api probe ran 9 of 9 cuLaunchKernel tests in bare, current and patched configurations

 The hardware tests were performed on a Nvidia a100 80gb PCIe with driver: 580.65.06 and CUDA toolkit 12.9. 

AI support: AI was used for a double check of the change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of repeated dynamic symbol lookups using `RTLD_NEXT`.
  * Removed unnecessary per-thread lookup tracking that could affect symbol resolution.

* **Tests**
  * Added automated coverage verifying repeated lookups return consistent, valid results.
  * Added validation for both standard execution and execution with the virtual GPU library preloaded.

### DIFF
--- a/src/libvgpu.c
+++ b/src/libvgpu.c
@@ -25,7 +25,6 @@ void *vgpulib;
 
 pthread_once_t pre_cuinit_flag = PTHREAD_ONCE_INIT;
 pthread_once_t post_cuinit_flag = PTHREAD_ONCE_INIT;
-pthread_once_t dlsym_init_flag = PTHREAD_ONCE_INIT;
 
 /* pidfound is to enable core utilization, if we don't find hostpid in container, then we have no
  where to find its core utilization */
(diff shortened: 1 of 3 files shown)


archlitchi (approved):
/lgtm

PR #280, opened 2026-08-16
fix(cuda): preserve managed-memory state in cuPointerGetAttributes

### LINKED ISSUE #281: cuPointerGetAttributes overrides managed-memory results and processes outputs after errors
## What happened

HAMi-core handles `CU_POINTER_ATTRIBUTE_IS_MANAGED` differently in the two CUDA pointer-attribute APIs.

The `cuPointerGetAttribute` hook calls the original CUDA driver function and returns its result without changing the managed-memory value.

However, the `cuPointerGetAttributes` hook calls the original CUDA function and then changes every requested `CU_POINTER_ATTRIBUTE_IS_MANAGED` result to `0`.

The multi-attribute hook also processes output values without first checking whether the original CUDA call returned `CUDA_SUCCESS`.

This means:

- `cuPointerGetAttribute` and `cuPointerGetAttributes` can return different  managed-memory states for the same pointer.
- A pointer allocated with `cuMemAllocManaged` can be incorrectly reported as  unmanaged.
- HAMi-core can read or modify output buffers after the original CUDA function  returns an error.

## Relevant source code

The affected implementation is in:

```text
src/cuda/memory.c
```

Relevant functions:

```c
cuPointerGetAttribute(...)
cuPointerGetAttributes(...)
```

Source reviewed:

https://github.com/Project-HAMi/HAMi-core/blob/5496322f2fb3e71bf1eca014fba3c9bc59ab8ffd/src/cuda/memory.c#L267-L290

The current `cuPointerGetAttributes` implementation follows this flow:

```c
CUresult res = CUDA_OVERRIDE_CALL(
    cuda_library_entry,
    cuPointerGetAttributes,
    numAttributes,
    attributes,
    data,
    ptr
);

for (int cur = 0; cur < numAttributes; cur++) {
    if (attributes[cur] == CU_POINTER_ATTRIBUTE_IS_MANAGED) {
        *(int *)(data[cur]) = 0;
    }
}

return res;
```

The output-processing loop runs regardless of the value returned in `res`.

## What you expected to happen

HAMi-core should preserve the managed-memory state returned by the CUDA driver.

For the same CUDA pointer:

- `cuPointerGetAttribute` and `cuPointerGetAttributes` should return the same  value for `CU_POINTER _ATTRIBUTE_IS_MANAGED`.
- Memory allocated with `cuMemAllocManaged` should be reported as managed by  both APIs.
- If the original `cuPointerGetAttributes` call returns an error, HAMi-core  should immediately return that error.
- When the original CUDA call fails, HAMi-core should not read, write, or  otherwise process the output buffers.

The CUDA Driver API describes `CU_POINTER_ATTRIBUTE_IS_MANAGED` as indicating whether a pointer refers to managed memory:

https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__UNIFIED.html

## How to reproduce it

This issue was identified through source-code analysis. I have not yet run the reproducer on real GPU hardware.

A focused CUDA test can reproduce the successful-query behavior:

1. Initialize CUDA using `cuInit`.
2. Get a CUDA device.
3. Create a CUDA context.
4. Allocate memory using `cuMemAllocManaged`.
5. Query `CU_POINTER_ATTRIBUTE_IS_MANAGED` using   `cuPointerGetAttribute`.
6. Query the same attribute using `cuPointerGetAttributes`.
7. Compare the values returned by both APIs.
8. Run the program without HAMi-core as a control.
9. Run the same program with HAMi-core loaded through `LD_PRELOAD`.

Example control run:

```bash
./test_pointer_attribute_managed
```

Example HAMi-core run:

```bash
CUDA_DEVICE_MEMORY_LIMIT=1g \
LD_PRELOAD=/path/to/libvgpu.so \
./test_pointer_attribute_managed
```

## Expected result

Both API calls should succeed and report the allocation as managed:

```text
cuPointerGetAttribute:  IS_MANAGED=1
cuPointerGetAttributes: IS_MANAGED=1
```

The two returned values should be equal.

## Current source-level result

The current hook changes the multi-attribute result to `0`:

```text
cuPointerGetAttribute:  IS_MANAGED=1
cuPointerGetAttributes: IS_MANAGED=0
```

This produces inconsistent results for the same managed allocation.

## Error-path behavior

The original CUDA function can return a non-success `CUresult`.

When this happens, the hook currently continues into its output-processing loop. Depending on the requested attributes, the hook can read or modify output values that should not be processed after the failed call.

The error path should be tested separately by making the original CUDA function return a non-success result and confirming that HAMi-core:

- Returns the same error.
- Does not enter the output-processing loop.
- Does not read or modify the output buffers.
- Does not crash when an output argument is invalid.

An invalid CUDA pointer should not be used as the only error-path test because the CUDA driver can return success with default attribute values for some invalid pointers. A deterministic invalid-argument case or a stubbed driver
function would provide more reliable coverage.

## Root cause

`cuPointerGetAttributes` performs HAMi-core-specific post-processing without first checking the result of the original CUDA driver call.

It also contains an unconditional override for `CU_POINTER_ATTRIBUTE_IS_MANAGED`.

The single-attribute hook does not contain the same override, which causes the two APIs to behave differently.

## Proposed solution

Update `cuPointerGetAttributes` so that it:

1. Calls the original CUDA driver function.
2. Immediately returns if the result is not `CUDA_SUCCESS`.
3. Does not override `CU_POINTER_ATTRIBUTE_IS_MANAGED`.
4. Preserves the managed-memory value returned by the CUDA driver.
5. Keeps any required successful-result processing for other attributes    unchanged.

Example intended control flow:

```c
CUresult res = CUDA_OVERRIDE_CALL(
    cuda_library_entry,
    cuPointerGetAttributes,
    numAttributes,
    attributes,
    data,
    ptr
);

if (res != CUDA_SUCCESS) {
    return res;
}

/* Process only successful results when required. */

return res;
```

## Testing requirements

Add regression coverage for the following cases:

### Successful managed-memory query

- Allocate memory using `cuMemAllocManaged`.
- Query `CU_POINTER_ATTRIBUTE_IS_MANAGED` through
  `cuPointerGetAttribute`.
- Query the same attribute through `cuPointerGetAttributes`.
- Confirm that both calls succeed.
- Confirm that both APIs return the same value.
- Confirm that both APIs return `1`.

### Driver-error behavior

- Make the original `cuPointerGetAttributes` call return an error.
- Confirm that HAMi-core returns the same error.
- Confirm that the hook does not process output buffers after the error.
- Confirm that the error path does not crash.

### Control comparison

Run the successful managed-memory test:

- Without HAMi-core.
- With HAMi-core loaded using `LD_PRELOAD`.

Record both results.

## Acceptance criteria

- [ ] `cuPointerGetAttributes` no longer forces       `CU_POINTER_ATTRIBUTE_IS_MANAGED` to `0`.
- [ ] `cuPointerGetAttribute` and `cuPointerGetAttributes` return the same       managed-memory state for the same pointer.
- [ ] A pointer allocated with `cuMemAllocManaged` is reported as managed.
- [ ] A failed original CUDA call is returned immediately.
- [ ] Output values are not processed after a failed CUDA call.
- [ ] Successful-query regression coverage is added.
- [ ] Driver-error regression coverage is added.
- [ ] The new test passes without `LD_PRELOAD`.
- [ ] The new test passes with HAMi-core loaded through `LD_PRELOAD`.
- [ ] The required HAMi-core build completes successfully.
- [ ] Real-GPU validation results are recorded.

## Impact

Applications can call the single-attribute and multi-attribute CUDA APIs in different code paths.

Returning different managed-memory states for the same pointer can cause an application to make incorrect decisions about:

- Memory placement.
- Memory migration.
- Unified-memory behavior.
- Pointer handling.
- CUDA compatibility paths.

Processing output buffers after an error can also cause incorrect results or unsafe memory access.

## Affected components

Likely affected files and components:

```text
src/cuda/memory.c
test/test_pointer_attribute_managed.c
```

No unrelated scheduler, device-plugin, or Kubernetes behavior should need to change.

## Environment

- HAMi-core revision reviewed:
  `5496322f2fb3e71bf1eca014fba3c9bc59ab8ffd`
- HAMi version: Not applicable to this source-level finding
- GPU model: Not yet tested
- NVIDIA driver version: Not yet tested
- CUDA version: Not yet tested
- Operating system: Not yet tested
- Validation status: Identified through source-code analysis; real-GPU  validation is pending

asadjan4611:
## Summary

This PR fixes inconsistent managed-memory reporting between `cuPointerGetAttribute` and `cuPointerGetAttributes`.

Previously, `cuPointerGetAttributes` overwrote `CU_POINTER_ATTRIBUTE_IS_MANAGED` with `0` after calling the CUDA driver. It also processed output values even when the driver returned an error.

This change preserves the CUDA driver result and returns immediately when the underlying call fails.

## Root cause

The `cuPointerGetAttributes` hook unconditionally post-processed requested attributes after the original CUDA call. That caused it to force managed memory to appear unmanaged and to access output buffers after an error.

## Related issue

Fixes #281

## Validation

- [x] `make check-cuda-hook-consistency`
- [x] `git diff --check upstream/main...HEAD`
- [ ] `make build-in-docker` — blocked by Docker Hub network connectivity
- [ ] Control run of `test_pointer_attribute_managed` on a real GPU
- [ ] `LD_PRELOAD` run of `test_pointer_attribute_managed` on a real GPU
- [ ] Error-path regression test

## Remaining validation

- Add regression coverage for the driver-error path.
- Complete the required Docker build.
- Run and record real-GPU validation, including GPU model, NVIDIA driver,
  CUDA version, operating system, and control/preloaded test results.

## User-facing impact

Yes. This corrects CUDA pointer-attribute results for managed memory.

## AI assistance disclosure

Codex assisted with source review or drafting this PR
description. I reviewed the change and take responsibility for its content.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CUDA pointer attribute handling when underlying queries fail.
  - Corrected managed-memory attribute reporting so related pointer queries return consistent results.
  - Prevented incomplete or misleading attribute processing after an unsuccessful CUDA operation.
  - Improved reliability when querying pointer attributes across different CUDA contexts and memory types.

- **Tests**
  - Added regression coverage for managed CUDA memory, pointer attribute consistency, invalid contexts, and CUDA-version-specific behavior.

### DIFF
--- a/src/cuda/memory.c
+++ b/src/cuda/memory.c
@@ -275,16 +275,15 @@ CUresult cuPointerGetAttributes ( unsigned int  numAttributes, CUpointer_attribu
     LOG_DEBUG("cuPointGetAttribute data=%p ptr=%llx", data, ptr);
     ENSURE_RUNNING();
     CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,cuPointerGetAttributes,numAttributes,attributes,data,ptr);
+    if (res != CUDA_SUCCESS) {
+        return res;
+    }
     int cur=0;
     for (cur=0;cur<numAttributes;cur++){
         if (attributes[cur]==CU_POINTER_ATTRIBUTE_MEMORY_TYPE){
             int j = check_memory_type(ptr);
             //*(int *)(data[cur])=1;
             LOG_DEBUG("check result = %d %d",j,*(int *)(data[cur]));
-        }else{
-            if (attributes[cur]==CU_POINTER_ATTRIBUTE_IS_MANAGED){
-                *(int *)(data[cur])=0;    
-            }
         }
     }
     return res;
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,6 +1,8 @@
 
 find_package(CUDA REQUIRED)
 
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/..)
+
 set(TEST_CPP_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 set(TEST_TARGET_NAMES_LIST) 
 
--- a/test/test_pointer_attribute_managed.c
+++ b/test/test_pointer_attribute_managed.c
@@ -1,0 +1,62 @@
+#include <stdio.h>
+#include <cuda.h>
+
+#include "test_utils.h"
+
+// Regression test for the managed-memory state reported by the two CUDA
+// pointer-attribute APIs.
+//
+// cuPointerGetAttribute and cuPointerGetAttributes must report the same
+// CU_POINTER_ATTRIBUTE_IS_MANAGED value for the same pointer, and a
+// cuMemAllocManaged allocation must be reported as managed.
+//
+// Run under the HAMi-core hooks, e.g.:
+//   LD_PRELOAD=./libvgpu.so ./test/test_pointer_attribute_managed
+
+int main() {
+    CHECK_DRV_API(cuInit(0));
+
+    CUdevice device;
+    CHECK_DRV_API(cuDeviceGet(&device, TEST_DEVICE_ID));
+
+    CUcontext ctx;
+#if CUDA_VERSION >= 13000
+    CHECK_DRV_API(cuCtxCreate(&ctx, NULL, 0, device));
+#else
+    CHECK_DRV_API(cuCtxCreate(&ctx, 0, device));
+#endif
+
+    CUdeviceptr dptr;
+    CHECK_DRV_API(cuMemAllocManaged(&dptr, 1 << 20, CU_MEM_ATTACH_GLOBAL));
+
+    int managed_single = -1;
+    CHECK_DRV_API(cuPointerGetAttribute(&managed_single,
+        CU_POINTER_ATTRIBUTE_IS_MANAGED, dptr));
+
+    int managed_multi = -1;
+    CUpointer_attribute attributes[1] = {CU_POINTER_ATTRIBUTE_IS_MANAGED};
+    void *data[1] = {&managed_multi};
+    CHECK_DRV_API(cuPointerGetAttributes(1, attributes, data, dptr));
+
+    printf("IS_MANAGED: cuPointerGetAttribute=%d cuPointerGetAttributes=%d\n",
+        managed_single, managed_multi);
+
+    if (managed_single != managed_multi) {
+        fprintf(stderr,
+            "cuPointerGetAttribute and cuPointerGetAttributes disagree on "
+            "IS_MANAGED: %d vs %d\n", managed_single, managed_multi);
+        return -1;
+    }
+
+    if (managed_multi != 1) {
+        fprintf(stderr,
+            "managed allocation reported as not managed: IS_MANAGED=%d\n",
+            managed_multi);
+        return -1;
+    }
+
+    cuMemFree(dptr);
+    CHECK_DRV_API(cuCtxDestroy(ctx));
+    printf("test_pointer_attribute_managed passed\n");
+    return 0;
+}
--- a/test/test_pointer_attribute_managed.c
+++ b/test/test_pointer_attribute_managed.c
@@ -1,7 +1,7 @@
 #include <stdio.h>
 #include <cuda.h>
 
-#include "test_utils.h"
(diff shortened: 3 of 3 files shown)


archlitchi (approved):
/lgtm

PR #282, opened 2026-08-17
fix: exclude concurrent seqlock writers

### LINKED ISSUE #269: Seqlock in get_gpu_memory_usage has no writer-writer exclusion
## What happened

`get_gpu_memory_usage()` reads each process slot with a seqlock, and the reader side is correct — it waits for an even sequence, reads, fences, re-reads, and retries if the sequence moved.

But a seqlock only gives that guarantee if writers are mutually exclusive with each other, and here they aren't.

`allocator.c` serialises its own writers under `pthread_mutex_t mutex`, so allocation paths are fine relative to each other. `cuda/context.c` doesn't take that mutex at all:

- `cuDevicePrimaryCtxRetain` → `add_gpu_device_memory_usage(getpid(), dev, context_size, 0)`
- `cuDevicePrimaryCtxRelease_v2` → `rm_gpu_device_memory_usage(getpid(), dev, context_size, 0)`

Both go through the `pid == getpid()` fast path and write the same slot (`region_info.my_slot`) as the allocator paths. There's no global lock around the CUDA hooks either.

So with thread A in `cuMemAlloc` and thread B in `cuDevicePrimaryCtxRetain`, the sequence counter can go even while writes are still in flight:

| step | seqlock | state |
| --- | --- | --- |
| A: `fetch_add` | 0 → 1 | odd — write in progress, correct |
| B: `fetch_add` | 1 → 2 | **even — readers now believe no write is active** |
| A: writes `used[dev].total` | 2 | a reader sampling here sees even and accepts |
| B: writes `used[dev].context_size` | 2 | |
| A: `fetch_add` | 2 → 3 | |
| B: `fetch_add` | 3 → 4 | |

A reader that samples in that window gets `seq1 == seq2`, both even, and concludes it has a consistent snapshot while two writers are mid-update. The protocol degrades to no protection, silently.

Individual fields are `_Atomic` so nothing tears, but the invariant the seqlock exists to protect — that `total` and the `context_size`/`module_size`/`data_size` breakdown are read from the same point in time — is not upheld.

## Second, related race in the same lines

`ctx_activate[dev]` is a plain `int`, read-modify-written with no atomics and no lock:

```c
if (ctx_activate[dev] == 0) {
    add_gpu_device_memory_usage(getpid(), dev, context_size, 0);
}
if (context_size > 0) {
    ctx_activate[dev] = 1;
}
```

Two threads retaining the same primary context can both observe `0` and both add `context_size`, double-counting that process's memory. The release path has the mirror problem.

## What I expected

Either writers serialised against each other so the seqlock means what it says, or the seqlock removed if it isn't needed.

## How to reproduce

Any multi-threaded CUDA process that allocates while another thread retains or releases a primary context — normal behaviour for inference servers. It's a race, so it's timing-dependent rather than deterministic; the window is the few instructions between the paired `fetch_add`s.

## Possible directions

I didn't want to send a patch before you've picked a direction, since they have quite different costs:

1. Have `context.c` take the same mutex `allocator.c` uses. Smallest change, but it makes an allocator-local mutex into a de facto module lock, which is probably the wrong home for it.
2. Give the slot its own writer mutex, used by every path that writes it. Clearer ownership, one more lock on the alloc path.
3. Drop the seqlock. `get_gpu_memory_usage()` only reads `used[dev].total`, which is a single `_Atomic` load and can't tear — the seqlock isn't buying that reader anything today. It would matter for any future reader that needs `total` and the breakdown together, so this depends on whether you want that guarantee.

Happy to send a PR for whichever you prefer, and to add a stress test that runs concurrent alloc and context retain/release against a slot and asserts readers never observe an inconsistent snapshot.

Jjateen:
Fixes #269

- writers bumped the seqlock with a bare `fetch_add`, so two of them
  interleave and drive it even mid-write, and a reader then accepts a
  snapshot torn across fields
- a mutex cannot fix it: the slow paths write another process's slot, so
  the racing writers can be in different processes
- writers now take the slot with a CAS, even to odd; a waiting writer
  never publishes odd, so the window readers see is unchanged
- adds a GPU-free ctest with two writers and two readers

## Before and after

| | torn reads accepted under an even sequence |
| --- | --- |
| before | 4165 |
| after | 0 |

400,000 add/remove pairs across two threads. The counter ends on exactly
1,600,000, so every acquire paired with its release.

```bash
cmake --build build --target test_seqlock_writer_exclusion
./build/test/test_seqlock_writer_exclusion        # passes

git stash push -- src/multiprocess/multiprocess_memory_limit.c
cmake --build build --target test_seqlock_writer_exclusion
./build/test/test_seqlock_writer_exclusion        # fails
git stash pop
```

## Hardware

RTX 5060 Laptop, driver 580.178.04, CUDA 13.2, base commit 5496322.

`test_alloc`, `test_runtime_alloc`, `test_alloc_host` and `test_host_register`
pass with the patched `libvgpu.so` at `CUDA_DEVICE_MEMORY_LIMIT=2G`.
`postinit_owner_death` and `seqlock_writer_exclusion` pass under ctest.

Single consumer GPU, so no multi-device or multi-container coverage. The race is
in the shared-region accounting and reproduces without a GPU at all.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when multiple processes update memory limits concurrently.
  * Reduced the risk of inconsistent or incomplete memory-accounting updates under heavy contention.
  * Added recovery handling for stalled update operations.

* **Tests**
  * Added regression coverage for concurrent memory-limit access and writer exclusion.
  * Added validation for consistent snapshots, balanced accounting, and successful completion.
  * Added test resource coordination and timeout handling for more reliable test execution.

### DIFF
--- a/src/multiprocess/multiprocess_memory_limit.c
+++ b/src/multiprocess/multiprocess_memory_limit.c
@@ -280,6 +280,17 @@ size_t get_gpu_memory_monitor(const int dev) {
 }
 
 // Lock-free memory usage aggregation with seqlock for consistent snapshots
+static inline void seqlock_cpu_relax(void) {
+#if defined(__x86_64__) || defined(__i386__)
+    __asm__ __volatile__("pause" ::: "memory");
+#elif defined(__aarch64__)
+    __asm__ __volatile__("yield" ::: "memory");
+#else
+    __asm__ __volatile__("" ::: "memory");   /* barrier only, no pause opcode */
+#endif
+}
+
+
 size_t get_gpu_memory_usage(const int dev) {
     LOG_INFO("get_gpu_memory_usage_lockfree dev=%d", dev);
     ensure_initialized();
@@ -305,12 +316,9 @@ size_t get_gpu_memory_usage(const int dev) {
             while (seq1 & 1) {
                 // Exponential backoff to reduce contention
                 if (retry_count < 5) {
-                    // First 5 retries: just CPU pause (fast path)
-                    #if defined(__x86_64__) || defined(__i386__)
-                    __asm__ __volatile__("pause" ::: "memory");
-                    #elif defined(__aarch64__)
-                    __asm__ __volatile__("yield" ::: "memory");
-                    #endif
+                    // First 5 retries: just CPU pause (fast path).  Shared with
+                    // the writer so both sides get the portable fallback.
+                    seqlock_cpu_relax();
                 } else if (retry_count < 20) {
                     // Next 15 retries: 1μs delay
                     usleep(1);
@@ -462,22 +470,49 @@ uint64_t nvml_get_device_memory_usage(const int dev) {
  * never publishes an odd value, so the window in which readers see a write in
  * progress is no longer than it is today.
  */
-static inline void seqlock_cpu_relax(void) {
-#if defined(__x86_64__) || defined(__i386__)
-    __asm__ __volatile__("pause" ::: "memory");
-#elif defined(__aarch64__)
-    __asm__ __volatile__("yield" ::: "memory");
-#else
-    __asm__ __volatile__("" ::: "memory");   /* barrier only, no pause opcode */
-#endif
-}
-
 /* The critical section is a handful of atomic adds with no syscall in it, so a
  * real wait is nanoseconds. Escalate anyway so a waiter never sits on a core:
  * pause, then yield, then short sleeps. */
-#define SEQLOCK_SPIN_LIMIT   1000u     /* pause only */
-#define SEQLOCK_YIELD_LIMIT  11000u    /* then sched_yield */
-#define SEQLOCK_SLEEP_LIMIT  31000u    /* then 100us naps, indefinitely */
+#define SEQLOCK_SPIN_LIMIT     1000u    /* pause only */
+#define SEQLOCK_YIELD_LIMIT   11000u    /* then sched_yield */
+#define SEQLOCK_RECLAIM_AFTER 31000u    /* then start trying to recover */
+#define SEQLOCK_RECLAIM_EVERY 10000u    /* and retry about once a second */
+
+/* A holder killed between begin and end leaves the sequence odd for good, which
+ * blocks every later writer and every reader of that slot. Recovery has to prove
+ * the holder is gone rather than infer it from a timeout, because a live writer
+ * can be descheduled for longer than any threshold we could pick.
+ *
+ * The writer therefore records its pid in the slot while it holds the sequence,
+ * and this only resets the counter when that pid no longer exists. Taking
+ * lock_shrreg() serialises the reclaim so two waiters cannot both perform it.
+ *
+ * Known residual: a writer killed in the window between winning the CAS and
+ * storing its pid leaves the sequence odd with seq_owner still 0. That is
+ * treated as an unknown holder and never reclaimed, so the slot stays wedged as
+ * it would have before this change. Closing it means publishing the sequence
+ * and the owner in one atomic word, which changes how the reader interprets the
+ * counter. */
+static void seqlock_try_reclaim(shrreg_proc_slot_t* slot) {
+    lock_shrreg();
+    uint64_t seq = atomic_load_explicit(&slot->seqlock, memory_order_acquire);
+    int32_t owner = atomic_load_explicit(&slot->seq_owner, memory_order_relaxed);
+    if ((seq & 1) && owner > 0 && kill(owner, 0) == -1 && errno == ESRCH) {
+        LOG_WARN("seqlock owner %d is gone, reclaiming slot", owner);
+        /* the dead writer stopped partway through, so this slot's counters no
+         * longer describe anything and are cleared with the sequence */
+        int dev;
+        for (dev = 0; dev < CUDA_DEVICE_MAX_COUNT; dev++) {
+            atomic_store_explicit(&slot->used[dev].total, 0, memory_order_relaxed);
+            atomic_store_explicit(&slot->used[dev].context_size, 0, memory_order_relaxed);
+            atomic_store_explicit(&slot->used[dev].module_size, 0, memory_order_relaxed);
+            atomic_store_explicit(&slot->used[dev].data_size, 0, memory_order_relaxed);
+        }
+        atomic_store_explicit(&slot->seq_owner, 0, memory_order_relaxed);
+        atomic_store_explicit(&slot->seqlock, seq + 1, memory_order_release);
+    }
+    unlock_shrreg();
+}
 
 static inline void seqlock_write_begin(shrreg_proc_slot_t* slot) {
     uint64_t seq;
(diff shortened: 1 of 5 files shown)


test/test_seqlock_writer_exclusion.c
```diff
@@ -0,0 +1,136 @@
...
+    shrreg_proc_slot_t *slot = region_info.my_slot;
+    uint64_t final_seq   = atomic_load(&slot->seqlock);
+    uint64_t final_total = atomic_load(&slot->used[DEV].total);
+
+    printf("writers=2 readers=2 iterations=%d\n", ITERATIONS);
+    printf("odd sequence observations : %" PRId64 "\n", odd);
+    printf("torn reads                : %" PRId64 "\n", torn);
+    printf("final seqlock             : %lu (%s)\n", final_seq,
+           (final_seq & 1) ? "ODD - a writer did not release" : "even");
+    printf("final total               : %lu (expected 0)\n", final_total);
```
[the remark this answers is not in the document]
Jjateen: This is now fixed, both now use `PRIu64`.

src/multiprocess/multiprocess_memory_limit.c
```diff
@@ -448,6 +448,46 @@ uint64_t nvml_get_device_memory_usage(const int dev) {
...
+        if (atomic_compare_exchange_weak_explicit(
+                &slot->seqlock, &seq, seq + 1,
+                memory_order_acq_rel, memory_order_relaxed))
+            return;
+    }
+}
+
+static inline void seqlock_write_end(shrreg_proc_slot_t* slot) {
+    atomic_fetch_add_explicit(&slot->seqlock, 1, memory_order_release);
+}
```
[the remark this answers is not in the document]
Jjateen: true, but this doesn't fit the scope of this PR as it is pre-existing behaviour.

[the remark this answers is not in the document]
Jjateen: You can create a follow up issue, but it's again not in scope of this PR

src/multiprocess/multiprocess_memory_limit.c
```diff
@@ -448,6 +449,72 @@ uint64_t nvml_get_device_memory_usage(const int dev) {
...
+        seq = atomic_load_explicit(&slot->seqlock, memory_order_relaxed);
+        if (seq & 1) {                 /* another writer holds the slot */
+            if (spins < SEQLOCK_SPIN_LIMIT) {
+                seqlock_cpu_relax();
+            } else if (spins < SEQLOCK_YIELD_LIMIT) {
+                sched_yield();
+            } else if (spins < SEQLOCK_SLEEP_LIMIT) {
+                usleep(100);
+            } else {
+                /* Keep waiting. Reclaiming the slot here is not safe: a live
```
mesutoezdil: if a writer dies mid write, seq stays odd. every later writer waits here forever. the reset near line 1129 only runs when the same pid comes back. what covers a dead foreign writer? did you try kill -9 mid write?

src/multiprocess/multiprocess_memory_limit.c
```diff
@@ -448,6 +449,72 @@ uint64_t nvml_get_device_memory_usage(const int dev) {
...
+    __asm__ __volatile__("" ::: "memory");   /* barrier only, no pause opcode */
+#endif
+}
+
+/* The critical section is a handful of atomic adds with no syscall in it, so a
+ * real wait is nanoseconds. Escalate anyway so a waiter never sits on a core:
+ * pause, then yield, then short sleeps. */
+#define SEQLOCK_SPIN_LIMIT   1000u     /* pause only */
+#define SEQLOCK_YIELD_LIMIT  11000u    /* then sched_yield */
+#define SEQLOCK_SLEEP_LIMIT  31000u    /* then 100us naps, indefinitely */
```
mesutoezdil: this limit changes nothing. both branches sleep 100us. drop the limit or the extra branch.

src/multiprocess/multiprocess_memory_limit.c
```diff
@@ -448,6 +449,72 @@ uint64_t nvml_get_device_memory_usage(const int dev) {
...
+                 * while that writer is still storing, or turn an already
+                 * released even sequence back to odd with no holder. Proving
+                 * the holder is dead needs an owner field this structure does
+                 * not have. */
+                usleep(100);
+            }
+            spins++;
+            continue;
+        }
+        if (atomic_compare_exchange_weak_explicit(
```
mesutoezdil, line 543: why acq_rel on success? begin publishes nothing yet. would acquire not be enough?

src/multiprocess/multiprocess_memory_limit.c
```diff
@@ -448,6 +449,72 @@ uint64_t nvml_get_device_memory_usage(const int dev) {
...
+ * A process-local mutex is not sufficient here: the slow paths below write
+ * another process's slot, so two writers to the same slot can live in
+ * different processes. Exclusion therefore has to sit in the shared region
+ * itself, which is what the CAS provides.
+ *
+ * Only the writer that wins the even->odd transition enters; a waiting writer
+ * never publishes an odd value, so the window in which readers see a write in
+ * progress is no longer than it is today.
+ */
+static inline void seqlock_cpu_relax(void) {
```
mesutoezdil: get_gpu_memory_usage has this same pause block inline. reuse this helper there.

test/test_seqlock_writer_exclusion.c
```diff
@@ -0,0 +1,169 @@
...
+#define QUANTUM      4096u          /* every write moves total by this much */
+#define ITERATIONS   200000
+#define DEV          0
+
+static _Atomic int stop_readers;
+static _Atomic int readers_ready;
+static _Atomic int64_t torn_reads;
+static _Atomic int64_t odd_observations;
+
+static void *writer_thread(void *arg) {
```
mesutoezdil, line 31: the race story is cross process. this test is two threads in one process on the fast path. why not two forked processes hitting the slow path?

mesutoezdil (commented):
good writeup and repro. questions inline.

src/multiprocess/multiprocess_memory_limit.c
Jjateen: Tried it: SIGKILL mid write leaves seq odd and the next writer blocks forever, and the 1129 reset only covers the same pid coming back. Proper fix can be owner pid in the slot's spare unused[0] + a reclaim under lock_shrreg() gated on ESRCH?

src/multiprocess/multiprocess_memory_limit.c
Jjateen: this is dead code left behind when I dropped the reclaim. I'll remove it in next commit

src/multiprocess/multiprocess_memory_limit.c
Jjateen: yes, it should be enough, I'll update

src/multiprocess/multiprocess_memory_limit.c
Jjateen: sure

test/test_seqlock_writer_exclusion.c
Jjateen: I'll add forked children each writing every slot by pid to force the slow path.

mesutoezdil:
resolve conflicts pls

Jjateen:
> resolve conflicts pls

resolved

maverick-woo:
Thanks for the PR. Our initial review found a risk in `seqlock_try_reclaim()`: a writer may update another process's slot through the slow path. If that writer dies, the current recovery clears all memory counters in the target slot, even though the target process may still be alive. This silently undercounts its GPU memory and may allow it to exceed the configured limit. Please do not clear the target slot based only on the writer's liveness; the recovery path must preserve or safely reconstruct the target process's accounting.

maverick-woo:
The comments are too verbose. Please keep them concise and only explain non-obvious design decisions.

Jjateen:
> The comments are too verbose. Please keep them concise and only explain non-obvious design decisions.

trimmed down to only non-obvious comments.

archlitchi:
please resolve these conflicts

src/multiprocess/multiprocess_memory_limit.c
```diff
@@ -476,6 +484,76 @@ uint64_t nvml_get_device_memory_usage(const int dev) {
...
+                    seqlock_try_reclaim(slot);
+            }
+            spins++;
+            continue;
+        }
+        /* acquire, not acq_rel: nothing is published until write_end */
+        if (atomic_compare_exchange_weak_explicit(
+                &slot->seqlock, &seq, seq + 1,
+                memory_order_acquire, memory_order_relaxed)) {
+            atomic_store_explicit(&slot->seq_owner, getpid(), memory_order_relaxed);
```
mesutoezdil: kill between cas and seq_owner store, owner stays zero?

PR #286, opened 2026-08-19
fix: scale utilization watcher correction step linearly with SM count

### LINKED ISSUE #274: CUDA_DEVICE_SM_LIMIT barely throttles on high-SM GPUs: correction step in the utilization watcher scales quadratically with SM count
## Environment

- GPU: NVIDIA RTX PRO 6000 Blackwell Server Edition (188 SMs, maxThreadsPerSM=1536)
- Driver 580.173.02, CUDA 13, PyTorch 2.13
- HAMi-core: 5496322 (main at the time of testing)
- `CUDA_DEVICE_SM_LIMIT=50`, `GPU_CORE_UTILIZATION_POLICY=FORCE`

## What happens

With the limit set to 50, a sustained matmul loop sits at 99-100% SM utilization for minutes at a time. The limiter is definitely wired up - hostPid gets discovered, `util_switch=1`, the limit is cached as 50, and `rate_limiter()` does get called from `cuLaunchKernelEx`. It just never throttles.

One watcher sample per minute under sustained load:

```
t        userutil  g_cur_cuda_cores
m1       99        3,984,504
m2       99        1,746,792
m3       48        5,012,008   refilled
m4-m11   ~99       oscillates between 1.2M and 8.9M
m12      81        -4,096      limiter finally blocks, device util drops to 17%
m13      99        5,762,893   refilled
```

Time-averaged utilization ends up near 100%. And anything with idle gaps shorter than the drain time never reaches the blocking path at all.

## Cause

`delta()` scales the correction increment by `sm_num * sm_num`, but the token pool is only linear in SM count (`g_total_cuda_cores = sm_num * max_thread_per_sm * 32`). Per 120ms tick that leaves a step/pool ratio of `sm_num * diff / 81920`, so it grows with device size.

At 188 SMs, a large utilization error plus the acceleration branch works out to roughly 2M tokens per tick, which refills the ~9.2M pool in well under a second. Draining it under load takes minutes. What comes out is a bang-bang cycle whose ON phase runs at ~100% for minutes.

I assume the constants were tuned on much smaller GPUs, where the quadratic term hadn't taken over yet.

## Fix 

Make the increment linear in SM count. The ratio then becomes `diff / 81920` regardless of device size, so a small GPU and a 188-SM one get the same controller behavior. I'd also clamp a single step to a tenth of the pool, in case something bigger shows up later.

Tested on the same GPU and the same load, 20s samples:

```
t      device util  userutil  share
20s    53%          45        17,742
40s    40%          49        21,024
60s    40%          52        16,068
80s    58%          51        17,202
100s   29%          41        18,331
120s   50%          50        18,449
```

Converges on the configured 50% in about 20s and stays in a narrow band after that.

## Note

I opened this as #271 before and it got closed - my fault for not filing an issue first. Happy to resubmit the patch against this issue if the direction looks right, and I can run more scenarios on this hardware if that would help.

maymust-jonghyunlee:
Fixes #274

Supersedes #271, which was closed because the issue had not been filed first; #274 now documents the hardware, reproduction, and measurements.

## Problem

The utilization watcher regulates future CUDA kernel submissions from NVML utilization feedback: `delta()` computes a correction step each 120 ms tick, and kernel launches spend tokens from a pool of `g_total_cuda_cores = sm_num * max_thread_per_sm * 32`.

The correction step scales with `sm_num * sm_num`, while the pool scales linearly with `sm_num`:

```text
pool       ∝ SM
correction ∝ SM²

correction / pool ∝ SM
```

So the relative controller gain grows with the device SM count. On an RTX PRO 6000 Blackwell Server Edition (188 SMs, driver 580.173.02, CUDA 13) with `CUDA_DEVICE_SM_LIMIT=50` and `GPU_CORE_UTILIZATION_POLICY=FORCE`, a large utilization error refilled the ~9.2M-token pool in under a second while draining it under sustained load took minutes — the controller degenerated into a bang-bang cycle whose ON phase ran at 99-100% utilization for minutes. The limiter itself was engaged (hostPid discovered, limit cached, `rate_limiter()` reached from `cuLaunchKernelEx`); the problem was the gain. Full traces are in #274.

## Fix

Drop one `sm_num` factor so the correction step scales like the pool:

```text
pool       ∝ SM
correction ∝ SM

correction / pool = utilization_diff / 81920
```

(before the existing acceleration multiplier) — independent of the device SM count, so small and large GPUs get the same controller dynamics.

Also promote the `g_total_cuda_cores` multiplication to `int64_t` so the intermediate `int` product cannot overflow before the assignment to the `int64_t` destination (raised during review of #271). This is defensive only — no overflow occurs on current devices (the largest value here today is ~9.2M vs INT_MAX ~2.1B).

Compared to #271, the extra `g_total_cuda_cores / 10` per-step cap was dropped to keep the diff minimal: after linear normalization the pre-acceleration step is bounded by `utilization_diff / 81920` of the pool (≤ ~0.13% per tick), and the cap never engaged in the recorded hardware runs.

## Validation

Reproduced and measured on real hardware as documented in #274 (RTX PRO 6000 Blackwell Server Edition, 188 SMs, driver 580.173.02, CUDA 13, PyTorch 2.13).

Before: sustained matmul load stayed near 99-100% utilization for minutes at a time despite the configured 50% limit.

With the linear correction scaling, measured utilization converged around the configured 50% target within roughly 20 seconds and stayed in a narrow band (20 s samples in #274: userutil 45 / 49 / 52 / 51 / 41 / 50).

`make build-in-docker` passes.

Two additional A/B runs (240 s sustained matmul, both libs built from the same tree, standalone LD_PRELOAD):

- `CUDA_DEVICE_SM_LIMIT=10` on the same 188-SM device: main averages ~26% device utilization with userutil swinging 0-72; patched converges to a 9-12 band within ~20 s and averages ~11%.
- `CUDA_DEVICE_SM_LIMIT=50` on a GTX 1080 (20 SMs): main degenerates into sustained 100% phases and averages ~81%; patched holds a band around 50 from the first ~20 s and averages ~49%.

Convergence takes roughly the same ~20 s on 20 SMs and 188 SMs — the normalized controller dynamics carry across a 9.4x SM range.

This PR intentionally does not change the kernel-cost model, the sampling interval, launch-hook behavior, or any other limiter semantics.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved GPU utilization correction across devices with different numbers of streaming multiprocessors.
  - Fixed arithmetic handling during CUDA core pool initialization to improve accuracy and reliability on larger configurations.
  - Enhanced consistency of GPU resource calculations across a wider range of hardware configurations.

### DIFF
--- a/src/multiprocess/multiprocess_utilization_watcher.c
+++ b/src/multiprocess/multiprocess_utilization_watcher.c
@@ -86,8 +86,11 @@ static void change_token(int64_t delta, int device_id) {
 static int64_t delta(int up_limit, int user_current, int64_t share, int device_id) {
   int utilization_diff =
       abs(up_limit - user_current) < 5 ? 5 : abs(up_limit - user_current);
+  /* Keep the correction step linear in the SM count so it scales like the
+   * token pool (sm_num * max_thread_per_sm * FACTOR) and the step-to-pool
+   * ratio is the same on every device size. */
   int64_t increment =
-      (int64_t)g_sm_num[device_id] * (int64_t)g_sm_num[device_id] *
+      (int64_t)g_sm_num[device_id] *
       (int64_t)g_max_thread_per_sm[device_id] * (int64_t)utilization_diff / 2560;
 
   /* Accelerate cuda cores allocation when utilization vary widely */
@@ -190,7 +193,7 @@ int setspec() {
             CU_DEVICE_ATTRIBUTE_MULTIPROCESSOR_COUNT, cu_dev));
         CHECK_CU_RESULT(cuDeviceGetAttribute(&g_max_thread_per_sm[dev],
             CU_DEVICE_ATTRIBUTE_MAX_THREADS_PER_MULTIPROCESSOR, cu_dev));
-        g_total_cuda_cores[dev] = g_max_thread_per_sm[dev] * g_sm_num[dev] * FACTOR;
+        g_total_cuda_cores[dev] = (int64_t)g_max_thread_per_sm[dev] * g_sm_num[dev] * FACTOR;
         LOG_INFO("setspec: device %d sm_num=%d max_threads_per_sm=%d total_cores=%ld FACTOR=%d",
                  dev, g_sm_num[dev], g_max_thread_per_sm[dev], g_total_cuda_cores[dev], FACTOR);
     }
--- a/src/multiprocess/multiprocess_utilization_watcher.c
+++ b/src/multiprocess/multiprocess_utilization_watcher.c
@@ -88,7 +88,9 @@ static int64_t delta(int up_limit, int user_current, int64_t share, int device_i
       abs(up_limit - user_current) < 5 ? 5 : abs(up_limit - user_current);
   /* Keep the correction step linear in the SM count so it scales like the
    * token pool (sm_num * max_thread_per_sm * FACTOR) and the step-to-pool
-   * ratio is the same on every device size. */
+   * ratio is the same on every device size:
+   *   increment / pool = utilization_diff / (2560 * FACTOR) = diff / 81920
+   * per 120 ms tick, before the acceleration branch below. */
   int64_t increment =
       (int64_t)g_sm_num[device_id] *
       (int64_t)g_max_thread_per_sm[device_id] * (int64_t)utilization_diff / 2560;


src/multiprocess/multiprocess_utilization_watcher.c
```diff
@@ -86,8 +86,11 @@ static void change_token(int64_t delta, int device_id) {
 static int64_t delta(int up_limit, int user_current, int64_t share, int device_id) {
   int utilization_diff =
       abs(up_limit - user_current) < 5 ? 5 : abs(up_limit - user_current);
+  /* Keep the correction step linear in the SM count so it scales like the
+   * token pool (sm_num * max_thread_per_sm * FACTOR) and the step-to-pool
+   * ratio is the same on every device size. */
   int64_t increment =
-      (int64_t)g_sm_num[device_id] * (int64_t)g_sm_num[device_id] *
+      (int64_t)g_sm_num[device_id] *
```
mesutoezdil, line 95: this makes the step smaller on small gpus. old step to pool ratio was sm_num * diff / 81920. new ratio is diff / 81920 on every gpu. so a 10 sm gpu now refills 10x slower than before. did you test on a small gpu? how many seconds to converge there? please paste the terminal output.

src/multiprocess/multiprocess_utilization_watcher.c
```diff
@@ -86,8 +86,11 @@ static void change_token(int64_t delta, int device_id) {
...
       (int64_t)g_max_thread_per_sm[device_id] * (int64_t)utilization_diff / 2560;
```
mesutoezdil, line 96: why keep 2560 here? after your change the real ratio is diff / (2560 * FACTOR). what is FACTOR in this repo and where is it defined? please add the final formula to the comment above so the next reader does not have to derive it.

src/multiprocess/multiprocess_utilization_watcher.c
mesutoezdil, line 95: the accel branch below multiplies the step by diff * 2 / (up_limit + 1). with a low limit like 10 that can be near 18x. did you check the step stays sane for low limits after this change, or only for limit 50?

src/multiprocess/multiprocess_utilization_watcher.c
```diff
@@ -190,7 +193,7 @@ int setspec() {
             CU_DEVICE_ATTRIBUTE_MULTIPROCESSOR_COUNT, cu_dev));
         CHECK_CU_RESULT(cuDeviceGetAttribute(&g_max_thread_per_sm[dev],
             CU_DEVICE_ATTRIBUTE_MAX_THREADS_PER_MULTIPROCESSOR, cu_dev));
-        g_total_cuda_cores[dev] = g_max_thread_per_sm[dev] * g_sm_num[dev] * FACTOR;
+        g_total_cuda_cores[dev] = (int64_t)g_max_thread_per_sm[dev] * g_sm_num[dev] * FACTOR;
```
mesutoezdil, line 198: the cast is fine but the old int math peaks near 9.2m on 188 sms. int holds up to about 2.1b. did you see a real overflow or is this only for safety? if only safety, please say so in the pr text.

mesutoezdil (commented):
thanks for the pr. the direction of the math looks right. i left questions inline. main ask: numbers from a small gpu. big gpu numbers are already in #274, small gpu is missing.

src/multiprocess/multiprocess_utilization_watcher.c
maymust-jonghyunlee: measured it on a small gpu: GTX 1080 (Pascal, 20 SMs, max_threads_per_sm=2048, pool=1,310,720), driver 580.173.02, `CUDA_DEVICE_SM_LIMIT=50` + `GPU_CORE_UTILIZATION_POLICY=FORCE`, sustained matmul, standalone LD_PRELOAD, both libs built from the same tree with the same toolchain (main 5496322 vs main + this patch). 240 s each, watcher samples every ~20 s:

main (5496322): userutil 46, 62, 9, 100, 17, 83, 95, 100, 100, 100, 100, 47 — oscillates, then sits pinned at 100 for the last ~80 s of the run, share slamming between 0 and ~95K. device util (nvidia-smi, 5 s samples) averages 81% against the 50 target.

patched: userutil 40, 41, 43, 55, 56, 68, 51, 67, 50, 43, 42 — in the target band from the first ~20 s sample onward, share steady around 9.5-10.2K. device util averages 49%.

so to answer directly: convergence on the 20-SM device takes about the same ~20 s as on the 188-SM one — with step and pool both linear in SM, the normalized trajectory carries over. and the old sm_num^2 gain wasn't a working small-GPU baseline to preserve: even at 20 SMs it degenerates into sustained 100% phases and averages 81% against a 50 limit.

the 10x is real relative to the old gain on the same device. if a hotter controller is ever wanted, a uniform multiplier on top of this normalization would raise it for every device alike — but i'd want to measure at that level before proposing a number. raw watcher logs from both runs available if you want them.

src/multiprocess/multiprocess_utilization_watcher.c
maymust-jonghyunlee: FACTOR is 32, defined in src/multiprocess/multiprocess_memory_limit.h. final ratio is utilization_diff / (2560 * 32) = diff / 81920 of the pool per 120 ms tick, before the accel branch. kept 2560 as-is to leave the existing tuning constant alone — the only intended change to this expression is dropping the sm_num factor. added the formula to the comment.

side note: the step is now algebraically g_total_cuda_cores * diff / 81920 (identical under integer division, before any pool doubling), so it could be derived from the pool directly — that would also make it follow the pool-doubling escape in utilization_watcher(), which the step ignores both before and after this patch. i left that out to keep this PR to the gain bug; can do it as a follow-up if the maintainers want it.

src/multiprocess/multiprocess_utilization_watcher.c
maymust-jonghyunlee: worked the worst cases on paper, then measured limit=10 on the same 188-SM box (standalone LD_PRELOAD run, same sustained matmul, both libs built from the same tree with the same toolchain — main 5496322 vs main + this patch).

paper worst case for the step as a fraction of the pool after the accel branch, 2 * diff^2 / (81920 * (up_limit + 1)):

| limit | max diff | this patch | main @ 188 SM |
|---|---|---|---|
| 50 | 50 | 0.12%/tick | ~22.5%/tick |
| 10 | 90 | 1.8%/tick | ~338%/tick (saturates in one tick) |
| 1 | 99 | ~12%/tick | ~2250%/tick |

measured, `CUDA_DEVICE_SM_LIMIT=10`, 240 s, watcher samples every ~20 s:

main (5496322): userutil 43, 22, 0, 25, 43, 28, 25, 31, 52, 27, 37 — swings between 0 and 72 the whole run, share slams between 0 and ~0.77M. device util (nvidia-smi, 5 s samples) averages 26% against the 10 target.

patched: userutil 9, 9, 10, 11, 10, 10, 10, 11, 12, 9, 12 — converges in ~20 s and stays in a 9-12 band. device util averages 11%.

so low limits were the worst case for the old scaling — the sm_num factor multiplies exactly this path — and the accel branch stays sane after the patch; share is clamped to [0, pool] in delta() either way. raw watcher logs available if you want them.

src/multiprocess/multiprocess_utilization_watcher.c
maymust-jonghyunlee: safety only, no overflow observed — the largest value here today is 1536 * 188 * 32 ≈ 9.2M, about 0.4% of INT_MAX; it would take sm_num * max_thread_per_sm > ~67M to actually overflow. it came out of review on #271 so i kept it. added a line to the PR text saying it's defensive.

maymust-jonghyunlee:
[hami-pr286-watcher-logs.zip](https://github.com/user-attachments/files/31248824/hami-pr286-watcher-logs.zip)

raw watcher logs for the runs quoted in the review threads:

- `gtx1080-20sm-limit50-*` — GTX 1080 (20 SMs), `CUDA_DEVICE_SM_LIMIT=50`, main vs patched
- `rtxpro6000-188sm-limit10-*` — RTX PRO 6000 (188 SMs), `CUDA_DEVICE_SM_LIMIT=10`, main vs patched

`*-watcher.log` is the burn process stderr with `LIBCUDA_LOG_LEVEL=3` — one `device 0: userutil=... share=...` line per 120 ms watcher tick, plus `setspec:` lines showing the detected sm_num / max_threads_per_sm / pool. `*-nvidia-smi.log` are host-side `utilization.gpu` samples every 5 s. README.md inside has the exact env, commands, and md5 of both libvgpu.so builds (same tree, same toolchain — main 5496322 vs this PR).

mesutoezdil:

PR #288, opened 2026-08-21
chore: remove obsolete unified lock and add PID discovery tests

chidwipak:
**What type of PR is this?**

Cleanup / Refactoring / Tests

**What this PR does / why we need it**:

Removes the unused `try_lock_unified_lock` and `try_unlock_unified_lock` functions from `src/utils.c` and their declarations from `src/include/utils.h`. These have been dead obsolete code since commit 7970f7f removed their usage. Removing them now prevents confusion around the active initialization locking path.

It also adds GPU-free regression tests (`test/test_pid_discovery.c`) for the host PID discovery helpers (`mergepid` and `getextrapid`) to ensure their boundary conditions and underflow behavior (fixed upstream in PR #265) remain correct during future optimizations.

**Which issue(s) this PR fixes**:

N/A (Routine cleanup and test coverage)

**Special notes for your reviewer**:

The new tests are registered in `test/CMakeLists.txt` using the established GPU-free methodology (linking only `-lrt -lpthread`) so they can be built and run in standard CI environments without requiring NVIDIA driver headers.

AI assistance: I used GitHub Copilot for code navigation and for understanding specific parts of the codebase while preparing this contribution. I reviewed and verified the changes and tests myself.

**Does this PR introduce a user-facing change?**:

No. This only removes dead code and adds internal regression tests.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved process and PID discovery handling for empty, boundary, underflow, and duplicate scenarios.
  - Removed obsolete locking behavior to simplify runtime coordination and reduce potential locking issues.
  - Consolidated process information definitions and declarations for more consistent behavior.

- **Tests**
  - Added automated coverage for PID discovery and duplicate-process handling.
  - Added validation for boundary and underflow conditions to improve reliability.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -14,26 +14,18 @@ foreach(TEST_SCRIPT ${TEST_SCRIPTS})
     get_filename_component(TEST_TARGET_NAME ${RELATIVE_TEST_PATH} NAME_WE)
     set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${TEST_TARGET_DIR})
 
-    if (TEST_TARGET_NAME STREQUAL "test_postinit_owner_death")
-        # Build this focused regression test with the production shared-region
-        # implementation.  It does not invoke any CUDA/NVML entry point at
-        # runtime.  Section garbage collection drops unrelated GPU-facing
-        # production functions.
-        add_executable(${TEST_TARGET_NAME}
-            ${TEST_SCRIPT}
-            ${CMAKE_CURRENT_SOURCE_DIR}/../src/multiprocess/multiprocess_memory_limit.c
-            ${CMAKE_CURRENT_SOURCE_DIR}/../src/log_utils.c)
-        target_compile_definitions(${TEST_TARGET_NAME} PRIVATE
-            _GNU_SOURCE)
-        target_compile_options(${TEST_TARGET_NAME} PRIVATE
-            -ffunction-sections -fdata-sections)
-        set_target_properties(${TEST_TARGET_NAME} PROPERTIES
-            LINK_FLAGS "-Wl,--no-export-dynamic,--gc-sections")
-    elseif (TEST_TARGET_NAME STREQUAL "test_pid_discovery")
-        add_executable(${TEST_TARGET_NAME}
-            ${TEST_SCRIPT}
-            ${CMAKE_CURRENT_SOURCE_DIR}/../src/utils.c
-            ${CMAKE_CURRENT_SOURCE_DIR}/../src/log_utils.c)
+    if (TEST_TARGET_NAME STREQUAL "test_postinit_owner_death" OR TEST_TARGET_NAME STREQUAL "test_pid_discovery")
+        if (TEST_TARGET_NAME STREQUAL "test_postinit_owner_death")
+            add_executable(${TEST_TARGET_NAME}
+                ${TEST_SCRIPT}
+                ${CMAKE_CURRENT_SOURCE_DIR}/../src/multiprocess/multiprocess_memory_limit.c
+                ${CMAKE_CURRENT_SOURCE_DIR}/../src/log_utils.c)
+        else()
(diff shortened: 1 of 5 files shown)


test/test_pid_discovery.c
```diff
@@ -0,0 +1,93 @@
...
+#define ASSERT_EQ(actual, expected) \
+    if ((actual) != (expected)) { \
+        fprintf(stderr, "%s:%d: Assertion failed: %d != %d\n", __FILE__, __LINE__, (int)(actual), (int)(expected)); \
+        exit(1); \
+    }
+
+void test_getextrapid_underflow() {
+    nvmlProcessInfo_t1 pre[] = { {100, 0}, {101, 0}, {102, 0} };
+    nvmlProcessInfo_t1 cur[] = { {100, 0} };
+    int extra = getextrapid(3, 1, pre, cur);
```
mesutoezdil, line 17: this passes on the old broken guard too. cur {100} exists in pre, so the loop also returns 0. use a pid not in pre, like {999}. old code returns 999, fixed code returns 0. only that pins the #265 fix.

test/test_pid_discovery.c
```diff
@@ -0,0 +1,93 @@
...
+#include <string.h>
+
+// Minimal types for mergepid and getextrapid
+typedef struct {
+    unsigned int pid;
+    unsigned long long usedGpuMemory;
+} nvmlProcessInfo_t1;
+
+// Forward declarations
+int mergepid(unsigned int *prev, unsigned int *current, nvmlProcessInfo_t1 *sub, nvmlProcessInfo_t1 *merged);
```
mesutoezdil: the struct and these prototypes are hand copies. a drift compiles fine and breaks at runtime. include nvml_override.h for the struct, or declare the functions in utils.h.

test/test_pid_discovery.c
```diff
@@ -0,0 +1,93 @@
...
+
+void test_getextrapid_happy_path() {
+    nvmlProcessInfo_t1 pre[] = { {100, 0} };
+    nvmlProcessInfo_t1 cur[] = { {100, 0}, {102, 0} };
+    int extra = getextrapid(1, 2, pre, cur);
+    ASSERT_EQ(extra, 102); // 102 is the new one
+}
+
+void test_getextrapid_empty() {
+    nvmlProcessInfo_t1 pre[] = {};
```
mesutoezdil: empty initializer is not standard c. pass NULL, prev is 0 anyway.

test/CMakeLists.txt
```diff
@@ -29,6 +29,17 @@ foreach(TEST_SCRIPT ${TEST_SCRIPTS})
             -ffunction-sections -fdata-sections)
         set_target_properties(${TEST_TARGET_NAME} PROPERTIES
             LINK_FLAGS "-Wl,--no-export-dynamic,--gc-sections")
+    elseif (TEST_TARGET_NAME STREQUAL "test_pid_discovery")
```
mesutoezdil: same flags as the postinit branch. fold them into one branch with OR, like the link block below.

mesutoezdil (changes requested):
see inline.

test/test_pid_discovery.c
chidwipak: I see the issue. Using {999, 0} makes the test catch the old bug properly. Fixed.

test/test_pid_discovery.c
chidwipak: Changed the empty initializer to NULL.

test/test_pid_discovery.c
chidwipak: Moved the struct and prototypes to utils.h so we don't have duplicate definitions in the test.

test/CMakeLists.txt
chidwipak: Combined the two test blocks into one OR condition to avoid the duplication.

maverick-woo:
Reviewed. LGTM, this can be merged. @archlitchi

archlitchi (approved):
/lgtm

Issue #276, opened 2026-08-13
HAMi-core Illegal device id → SIGSEGV crash on multi-GPU (torchrun) vGPU jobs


shubh3794:
[...]

Issue #234, opened 2026-07-27
[Bug/Question] SIGSEGV in cuMemGetInfo_v2 when running NVIDIA Isaac Sim / OptiX in vGPU container


liugs0213:
### Environment
- **Workload**: NVIDIA Isaac Sim 4.x / 5.x (with OptiX Denoiser enabled)
- **vGPU Mechanism**: API Interception (`libvgpu.so` hooking CUDA driver calls)

### Issue Description
When running NVIDIA Isaac Sim inside a vGPU-enabled container, the process crashes with `Segmentation fault (core dumped)` (`exit code 139`) shortly after `app ready` when OptiX denoiser (`librtx.optixdenoising.plugin.so`) is initialized.

The crash backtrace shows that `libvgpu.so` encounters a null pointer dereference inside `cuMemGetInfo_v2`:

```text
Thread backtrace:
000: libc.so.6!__sigaction+0x50
001: libvgpu.so!cuMemGetInfo_v2+0xd2              <-- Crash inside vGPU interception hook
002: libnvoptix.so.1!rtGetSymbolTable+0x14646d
003: librtx.optixdenoising.plugin.so!carbOnPluginPreStartup+0x3d9f
```

It seems OptiX calls `cuMemGetInfo_v2` from its internal rendering threads, which passes a context descriptor that the vGPU hook library fails to parse safely.

### Workaround
We mitigated this at the application level using an `LD_PRELOAD` C library that bypasses `libvgpu.so` for `cuMemGetInfo_v2` and forwards calls directly to host native `libcuda.so.1`:

```c
#define _GNU_SOURCE
#include <stdio.h>
#include <dlfcn.h>

typedef int (*pfn_cuMemGetInfo_v2)(size_t *free_mem, size_t *total_mem);

int cuMemGetInfo_v2(size_t *free_mem, size_t *total_mem) {
    static pfn_cuMemGetInfo_v2 real_fn = NULL;
    if (!real_fn) {
        void *h = dlopen("libcuda.so.1", RTLD_LAZY | RTLD_LOCAL);
        if (!h) h = dlopen("libcuda.so", RTLD_LAZY | RTLD_LOCAL);
        if (h) real_fn = (pfn_cuMemGetInfo_v2)dlsym(h, "cuMemGetInfo_v2");
    }
    if (real_fn) {
        int res = real_fn(free_mem, total_mem);
        if (res == 0 && free_mem && *free_mem > 0) return 0;
    }
[...]

PR #290, opened 2026-08-23
fix: prevent OOB SIGSEGV from an undefined device id on context-less threads

### LINKED ISSUE #289: Harden device-id handling: a context-less thread passes an undefined device id into per-device shared-region arrays, causing OOB access and SIGSEGV (root cause behind #276 and #234)
### Summary

On a thread that has no current CUDA context, `cuCtxGetDevice()` returns `CUDA_ERROR_INVALID_CONTEXT` and does not write its output parameter. The uninitialized value is then used as a device id to index the fixed-size `[CUDA_DEVICE_MAX_COUNT]` per-device arrays in the mmap'd shared region (`limit[]`, `sm_limit[]`, `used[]`, and so on). An out-of-range index is an out-of-bounds read or write that crashes the process with `SIGSEGV`.

This is the common root cause behind #276 and #234, which both report SIGSEGV crashes with large or nonsensical device ids in the logs.

### Where it happens

The bad device id reaches the per-device arrays from two kinds of sites:

1. Allocator source sites in `src/allocator/allocator.c` (`add_chunk`, `add_chunk_async`, `add_chunk_from_pool_async`, `remove_chunk_async`, `oom_check`) call `cuCtxGetDevice(&dev)` and use `dev` without checking the return code. On a context-less thread `dev` is undefined.
2. The shared-region accessors in `src/multiprocess/multiprocess_memory_limit.c` index the per-device arrays by that id.

Notably, five getters already have the guard condition but no early return, so they log the illegal id and then index the array anyway:

```c
uint64_t get_current_device_memory_limit(const int dev) {
    ensure_initialized();
    if (dev < 0 || dev >= CUDA_DEVICE_MAX_COUNT) {
        LOG_ERROR("Illegal device id: %d", dev);
        // no return here, so execution falls through
    }
    return region_info.shared_region->limit[dev];   // out of bounds when dev is illegal
}
```

`cuda_to_nvml_map()` in `src/multiprocess/multiprocess_utilization_watcher.c` also indexes its lookup array with no upper-bound check.

### When it manifests

Deterministically on multi-rank torchrun workers, where the main thread initializes a context but a worker thread performing an allocation never binds one. The undefined id is whatever happened to be on the worker stack. Field crash reports show ids such as `1716861768`.

### Reproduction (no GPU required)

Compile any of the accessors above with an out-of-range id. A minimal harness that calls `get_current_device_memory_limit(1716861768)` (and the other accessors) against current `main` crashes with SIGSEGV (exit 139). The same harness against a build that adds the missing bounds checks returns a safe value and exits cleanly. The harness also shows a silent out-of-bounds write first: `set_current_device_memory_limit(-1, 4096)` stores into `limit[-1]`, and a subsequent read returns `4096`.

### Proposed fix

Harden device-id handling in three places, none of which changes behavior for valid ids:

1. Allocator sources: on `cuCtxGetDevice` failure, skip per-device accounting or enforcement and forward the allocation to the driver rather than indexing per-device state with an undefined id. Never fall back to device 0.
2. Shared-region accessors: bounds-check the id and return a safe sentinel (-1 for the int helpers, 0 for the size and usage getters).
3. NVML mapping: `cuda_to_nvml_map()` rejects an out-of-range id and returns a sentinel.

Plus a GPU-free regression test that feeds a battery of out-of-range ids into every guarded accessor and asserts no crash.

I have a fix ready and will open a PR referencing this issue.

Eshiv-Pandey:
**What type of PR is this?**

Bug fix.

**What this PR does / why we need it**:

On a thread with no current CUDA context, `cuCtxGetDevice()` returns `CUDA_ERROR_INVALID_CONTEXT` and does not write its output parameter. The uninitialized value was then used as a device id to index the fixed-size `[CUDA_DEVICE_MAX_COUNT]` per-device arrays in the mmap'd shared region, an out-of-bounds access that crashes the process with SIGSEGV. This reproduces deterministically on multi-rank torchrun workers and is the root cause behind #276 and #234.

The fix hardens device-id handling in three layers, none of which changes behavior for valid ids:

1. Allocator sources (`src/allocator/allocator.c`): `add_chunk`, `add_chunk_async`, `add_chunk_from_pool_async`, and `oom_check` now check the `cuCtxGetDevice` return code. On failure they forward the allocation to the driver or skip enforcement rather than indexing per-device state with an undefined id. They never fall back to device 0. `remove_chunk_async` no longer consults the current context at all: it releases against the device recorded on the allocation entry, captured before `LIST_REMOVE` frees that entry, which stays correct even when the freeing thread has no context or a different current device.
2. Shared-region accessors (`src/multiprocess/multiprocess_memory_limit.c`): each accessor bounds-checks the id and returns a safe sentinel. Five getters already had the guard condition but were missing the early return, so they logged the illegal id and then indexed the array anyway; this adds the missing returns (-1 for the int helpers, 0 for the size and usage getters).
3. NVML mapping (`src/multiprocess/multiprocess_utilization_watcher.c`): `cuda_to_nvml_map()` rejects an out-of-range id and returns a sentinel instead of indexing its lookup array.

A GPU-free regression test (`test/test_device_id_guard.c`) feeds a battery of out-of-range ids, including the exact value seen in the field, into every guarded accessor and asserts a safe return with no crash, and confirms the valid in-range path is untouched.

**Which issue(s) this PR fixes**:

Fixes #289

**Special notes for your reviewer**:

Related to #276 and #234, which report the same SIGSEGV with large or nonsensical device ids.

Validation performed locally:
- CI-parity build in the nvidia/cuda:13.3.0-cudnn-devel-ubi8 image (matches the repo build workflow): clean, no new warnings.
- New `device_id_guard` test passes, as do `postinit_owner_death` and `dlsym_rtld_next`. `dlsym_rtld_next_preloaded` fails locally because the container has no NVIDIA driver and cannot load `libcuda.so.1`; it fails identically on a pristine `origin/main` build, so it is pre-existing and environmental.
- Hook static consistency check passes (entries 211, wrappers 217).
- cpplint (linelength 120) reports no findings on any changed line.
- Deterministic before/after: the same harness crashes with SIGSEGV (exit 139) against `main` and exits cleanly against this branch. It also demonstrates the silent out-of-bounds write on `main`, where `set_current_device_memory_limit(-1, 4096)` stores into `limit[-1]` and a subsequent read returns 4096.

I do not have multi-GPU hardware, so I could not add a live end-to-end torchrun repro; the regression test reproduces the crash path without a GPU.

Review follow-up: the second commit reworks `remove_chunk_async`. An earlier revision of this PR skipped the usage release when the freeing thread had no context, which would have left the allocation charged and produced a spurious OOM later. Thanks to the automated review for spotting that. The device has to be read before `LIST_REMOVE`, since that macro frees the entry.

One related observation I deliberately left alone, happy to fold it in if you would prefer: when an allocation is forwarded untracked, a later `cuMemFree_v2` reaches `remove_chunk`, which returns -1 for an unknown pointer. That path looks unreachable in practice, since a context-less `cuMemAlloc_v2` fails at the driver and nothing is allocated to free. The async free path already handles the untracked case by forwarding the free to the driver, so making the synchronous path symmetric would be a small follow-up.

AI assistance disclosure: I used AI (Claude) to help analyze the crash and draft this change. I reviewed, edited, and validated the code and wrote the commit messages myself.

**Does this PR introduce a user-facing change?**:

```release-note
Fix a SIGSEGV caused by an undefined device id from a context-less thread being used to index per-device shared-region arrays.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory allocation accounting to prevent concurrent allocations from exceeding device memory limits.
  * Ensured failed allocations roll back reserved capacity safely.
  * Improved handling of allocations when no CUDA context is active.
  * Ensured asynchronous memory removal uses the allocation’s device for accurate accounting.
  * Added safeguards for invalid GPU device IDs across memory tracking and monitoring operations.
  * Prevented unsafe device mapping access and provided appropriate error handling.

* **Tests**
  * Added regression coverage for concurrent allocation limits, process cleanup, device validation, and utilization monitoring.
  * Added GPU-free tests for invalid and valid device ID handling.

### DIFF
--- a/src/allocator/allocator.c
+++ b/src/allocator/allocator.c
@@ -316,10 +316,7 @@ int add_chunk_async(CUdeviceptr *address, size_t size, CUstream hStream) {
     size_t addr = 0;
     CUresult res = CUDA_SUCCESS;
     CUdevice dev;
-    if (cuCtxGetDevice(&dev) != CUDA_SUCCESS) {
-        LOG_WARN("add_chunk_async: no current context, forwarding allocation without tracking");
-        return CUDA_OVERRIDE_CALL(cuda_library_entry, cuMemAllocAsync, address, size, hStream);
-    }
+    cuCtxGetDevice(&dev);
     if (oom_check(dev, size))
         return CUDA_ERROR_OUT_OF_MEMORY;
 
@@ -345,10 +342,7 @@ int add_chunk_from_pool_async(CUdeviceptr *address, size_t size, CUmemoryPool po
     size_t addr = 0;
     CUresult res = CUDA_SUCCESS;
     CUdevice dev;
-    if (cuCtxGetDevice(&dev) != CUDA_SUCCESS) {
-        LOG_WARN("add_chunk_from_pool_async: no current context, forwarding allocation without tracking");
-        return CUDA_OVERRIDE_CALL(cuda_library_entry, cuMemAllocFromPoolAsync, address, size, pool, hStream);
-    }
+    cuCtxGetDevice(&dev);
     if (oom_check(dev, size))
         return CUDA_ERROR_OUT_OF_MEMORY;
 
--- a/src/allocator/allocator.c
+++ b/src/allocator/allocator.c
@@ -254,18 +254,22 @@ int free_raw(CUdeviceptr dptr) {
 int remove_chunk_async(
     allocated_list *a_list, CUdeviceptr dptr, CUstream hStream) {
     size_t t_size;
+    CUdevice t_dev;
     allocated_list_entry *val;
     for (val = a_list->head; val != NULL; val = val->next) {
         if (val->entry->address == dptr) {
             t_size=val->entry->length;
+            /* Release against the device recorded when the allocation was
+             * charged, captured before LIST_REMOVE frees the entry. The
+             * freeing thread may have no current context, or a different
+             * current device, so asking cuCtxGetDevice here would skip or
+             * misattribute the release and leave the usage charged, which
+             * later surfaces as a spurious OOM. */
+            t_dev = val->entry->dev;
             CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemFreeAsync,dptr,hStream);
             LIST_REMOVE(a_list,val);
             a_list->limit-=t_size;
-            CUdevice dev;
-            if (cuCtxGetDevice(&dev) == CUDA_SUCCESS)
-                rm_gpu_device_memory_usage(getpid(), dev, t_size, 2);
-            else
-                LOG_WARN("remove_chunk_async: no current context, skipping accounting");
+            rm_gpu_device_memory_usage(getpid(), t_dev, t_size, 2);
             return 0;
         }
     }
--- a/src/allocator/allocator.c
+++ b/src/allocator/allocator.c
@@ -35,10 +35,14 @@ size_t round_up(size_t size, size_t unit) {
 
 int oom_check(const int dev, size_t addon) {
     CUdevice d;
-    if (dev==-1)
-        cuCtxGetDevice(&d);
-    else
+    if (dev == -1) {
+        if (cuCtxGetDevice(&d) != CUDA_SUCCESS) {
+            LOG_WARN("oom_check: no current context, skipping enforcement");
+            return 0;
+        }
+    } else {
         d=dev;
+    }
     uint64_t limit = get_current_device_memory_limit(d);
     size_t _usage = get_gpu_memory_usage(d);
 
@@ -113,7 +117,17 @@ int add_chunk(CUdeviceptr *address, size_t size) {
     CUdevice dev;
     CUresult res;
 
-    cuCtxGetDevice(&dev);
+    if (cuCtxGetDevice(&dev) != CUDA_SUCCESS) {
+        /* No current context on this thread (for example a worker thread that
+         * never bound one). We cannot attribute or bound this allocation to a
+         * device, so forward it to the driver rather than index per-device
+         * state with an undefined id. */
+        LOG_WARN("add_chunk: no current context, forwarding allocation without tracking");
+        if (size <= IPCSIZE) {
+            return CUDA_OVERRIDE_CALL(cuda_library_entry, cuMemAlloc_v2, address, size);
+        }
+        return cuMemoryAllocate(address, size, NULL);
+    }
 
     /* OOM pre-check without lock */
     if (oom_check(dev, size))
@@ -248,8 +262,10 @@ int remove_chunk_async(
             LIST_REMOVE(a_list,val);
             a_list->limit-=t_size;
             CUdevice dev;
-            cuCtxGetDevice(&dev);
-            rm_gpu_device_memory_usage(getpid(),dev,t_size,2);
+            if (cuCtxGetDevice(&dev) == CUDA_SUCCESS)
+                rm_gpu_device_memory_usage(getpid(), dev, t_size, 2);
+            else
+                LOG_WARN("remove_chunk_async: no current context, skipping accounting");
             return 0;
         }
     }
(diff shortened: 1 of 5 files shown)


src/allocator/allocator.c
```diff
@@ -248,8 +262,10 @@ int remove_chunk_async(
             LIST_REMOVE(a_list,val);
             a_list->limit-=t_size;
             CUdevice dev;
-            cuCtxGetDevice(&dev);
-            rm_gpu_device_memory_usage(getpid(),dev,t_size,2);
+            if (cuCtxGetDevice(&dev) == CUDA_SUCCESS)
+                rm_gpu_device_memory_usage(getpid(), dev, t_size, 2);
+            else
+                LOG_WARN("remove_chunk_async: no current context, skipping accounting");
```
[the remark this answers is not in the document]
Eshiv-Pandey: Good catch, fixed in ee875ef. You are right that the earlier revision left the usage charged: the allocation was charged at alloc time, so skipping the release on a context-less free leaks accounting and shows up later as a spurious OOM.

One adjustment to the suggested diff: `LIST_REMOVE` does `free(val->entry); free(val);`, so reading `val->entry->dev` after it would be a use-after-free. I capture the device before `LIST_REMOVE` instead, which is the pattern `remove_chunk` already uses on the synchronous path:

```c
t_size = val->entry->length;
t_dev  = val->entry->dev;   /* before LIST_REMOVE frees the entry */
CUDA_OVERRIDE_CALL(cuda_library_entry, cuMemFreeAsync, dptr, hStream);
LIST_REMOVE(a_list, val);
a_list->limit -= t_size;
rm_gpu_device_memory_usage(getpid(), t_dev, t_size, 2);
```

This also drops the `cuCtxGetDevice` query from the free path entirely, so a thread freeing memory that was allocated on a different device now un-charges the correct one.

maverick-woo:
Thanks @Eshiv-Pandey for the PR. It fixes the invalid device index problems in #276 and #289.

The async allocation changes are not needed for these two issues. Please move the changes to `add_chunk_async` and `add_chunk_from_pool_async` to a follow-up PR, so this PR stays focused.

For #234, please keep it open for now. Its `cuMemGetInfo_v2`/OptiX path is different and needs separate testing. My suggestion is to check the return value of the real `cuMemGetInfo_v2` before using `free` and `total`.

Eshiv-Pandey:
Thanks @maverick-woo.

-> async paths: reverted the `add_chunk_async` and `add_chunk_from_pool_async` changes to keep this PR focused on #276 and #289. I will open a follow-up PR for the context-less guards on those paths.
-> #234: kept open. Agreed the `cuMemGetInfo_v2`/OptiX path is separate; the follow-up will check the real `cuMemGetInfo_v2` return value before using `free`/`total`, with its own testing.

happy to work on the two follow ups :)

nishitnshah:
We are also blocked on this PR. Any update?

maverick-woo:
Sorry for the delay. We'll get this reviewed as soon as possible. @Eshiv-Pandey, please resolve the current conflicts. Thanks!

src/allocator/allocator.c
```diff
@@ -240,16 +254,22 @@ int free_raw(CUdeviceptr dptr) {
...
     for (val = a_list->head; val != NULL; val = val->next) {
         if (val->entry->address == dptr) {
             t_size=val->entry->length;
+            /* Release against the device recorded when the allocation was
+             * charged, captured before LIST_REMOVE frees the entry. The
+             * freeing thread may have no current context, or a different
+             * current device, so asking cuCtxGetDevice here would skip or
+             * misattribute the release and leave the usage charged, which
+             * later surfaces as a spurious OOM. */
+            t_dev = val->entry->dev;
```
mesutoezdil, line 302: #311 makes this exact change on its own. worth agreeing which of the two carries it?

Eshiv-Pandey: Sorted, Ayush has agreed for keeping this in #290 only.

archlitchi (approved):
/lgtm

PR #291, opened 2026-08-23
fix: make ctx_activate carry its own 0 <-> 1 transition

AshrafAhmed9:
Second race from #269 — #282 covers the seqlock, not this one. I asked on the
issue whether you wanted a new issue or that one reopened and didn't hear back,
so here's the patch.

`cuDevicePrimaryCtxRetain` tested the flag and set it in two separate
statements:

```c
if (ctx_activate[dev] == 0) {
    add_gpu_device_memory_usage(getpid(),dev,context_size,0);
}
if (context_size>0) {
    ctx_activate[dev] = 1;
}
```

Nothing holds across those two ifs. Two threads retaining the same primary
context both read 0, both call `add_gpu_device_memory_usage`, and the process
gets charged twice for one context. Release has the same problem in reverse.

This one is process-local so it doesn't need the cross-process CAS from #282.
`_Atomic int` and a compare-exchange covers it — whoever wins 0 -> 1 does the
add, whoever wins 1 -> 0 does the remove.

I moved the declaration into the header behind
`ctx_activate_acquire`/`ctx_activate_release`, mostly so the test can call the
real thing. `context.c` won't link without the CUDA hooks, so a test that
included it couldn't build GPU-free.

While doing that I noticed `context.c` declares the array as
`extern int ctx_activate[16]` while `multiprocess_memory_limit.c` defines it as
`[32]`. Not compatible types across TUs. Both are `CUDA_DEVICE_MAX_COUNT` now,
which is what `dev` already indexes `used[]` with a line later.

One behaviour change worth flagging: the old code called
`add_gpu_device_memory_usage(..., context_size, 0)` even when `context_size`
was 0, which added nothing and left the flag clear. The new code skips the call.

Four threads, 200k iterations each, built in the CI image:

| | adds | removes | drift |
| --- | --- | --- | --- |
| before | 734,980 | 734,964 | 16 |
| before | 727,138 | 727,180 | -42 |
| before | 759,907 | 759,938 | -31 |
| after | 209,146 | 209,146 | 0 |

Drift is accounting the process never gets back, or a release for a charge it
never made. Small per run, but it doesn't wash out — a long-lived process keeps
collecting it. Five runs patched, all zero.

```bash
cmake --build build --target test_ctx_activate_race
./build/test/test_ctx_activate_race
```

To see it fail, put the two helpers back to a plain `int` with a separate
test-then-set and rebuild.

`make build-in-docker` is clean. Under ctest `ctx_activate_race`,
`postinit_owner_death` and `dlsym_rtld_next` pass. `dlsym_rtld_next_preloaded`
fails in my container because it LD_PRELOADs libvgpu and there's no driver
there. No multi-GPU box here either, but the race is in process-local
accounting and repros without a GPU at all.

Didn't touch anything else in `context.c`. It still doesn't check `res` from
the retain before accounting, which seemed like a separate question.

### DIFF
--- a/src/cuda/context.c
+++ b/src/cuda/context.c
@@ -2,7 +2,6 @@
 #include "multiprocess/multiprocess_memory_limit.h"
 
 extern size_t context_size;
-extern int ctx_activate[16];
 
 
 CUresult cuDevicePrimaryCtxGetState( CUdevice dev, unsigned int* flags, int* active ){
@@ -15,11 +14,8 @@ CUresult cuDevicePrimaryCtxRetain(CUcontext *pctx, CUdevice dev){
     LOG_INFO("dev=%d context_size=%ld",dev,context_size);
     //for Initialization only
     CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,cuDevicePrimaryCtxRetain,pctx,dev);
-    if (ctx_activate[dev] == 0) {
-        add_gpu_device_memory_usage(getpid(),dev,context_size,0); 
-    }
-    if (context_size>0) {
-        ctx_activate[dev] = 1;
+    if (context_size > 0 && ctx_activate_acquire(dev)) {
+        add_gpu_device_memory_usage(getpid(), dev, context_size, 0);
     }
     return res;
 }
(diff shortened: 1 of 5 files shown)


AshrafAhmed9:
cpplint whitespace fixed.

On the flag changing before the charge lands — that's right, and it did change.
The old retain charged and then set the flag, this one sets the flag and then
charges. So a release can win 1 -> 0 and call `rm_gpu_device_memory_usage`
before the add has run.

To actually hit it the app has to release a primary context on one thread while
another thread is still inside the retain for the same device. That's already
broken refcounting on the CUDA side, and the driver call underneath would be
unbalanced too.

It's also transient rather than lost. `total` is `_Atomic uint64_t` with
fetch_add/fetch_sub, so the two commute and it converges once both land. A
reader in that gap sees a bad number for the length of one atomic op, not the
permanent drift this PR is fixing.

Happy to close it anyway if you'd rather: CAS 0 -> 2, charge, store 1 on
retain, and CAS 1 -> 2, uncharge, store 0 on release, so nothing can win the
transition while a charge is in flight. Ten lines or so. I left it out because
it only matters under API misuse and I didn't want to grow the patch past the
bug, but say the word.

mesutoezdil:

PR #292, opened 2026-08-25
Fix inherited exit cleanup in clone children

Liauuu:
A child process created without running the `pthread_atfork` child handler can inherit the parent's initialized `region_info`, including `region_info.pid`.

For example, a child created with direct `clone()` does not run `child_reinit_flag`, so both the cached parent PID and initialized `region_info.init_status` remain in the child. If the child then exits with `exit()` without re-initializing, the inherited `exit_handler` runs.

Previously, `exit_handler` used:

```c id="82x1ue"
int32_t my_pid = region_info.pid;
```

Since `region_info.pid` can still contain the parent's PID, the child may treat the parent's state as its own. It can call `sem_post()` on a semaphore still owned by the parent and clear the parent's entry in `procs[]`. This could allow another process to enter the critical section or remove the parent's GPU usage/accounting state.

**Fix**

I changed `exit_handler` to use the actual PID of the running process:

```c id="zqkk11"
int32_t my_pid = getpid();
```

If the cached `region_info.pid` is set but does not match `getpid()`, the handler returns without modifying the inherited state.

I also changed `lock_shrreg()` to record `owner_pid` using `getpid()`, so the recorded lock owner is always the actual process holding the lock.

For the normal `fork()` path, `child_reinit_flag` already resets:

```c id="y45sw8"
region_info.init_status = PTHREAD_ONCE_INIT;
```

I additionally update `region_info.pid` with `getpid()` there, keeping the cached PID consistent for later initialization or shared-region work in the child.

**Test**

Tested on Ubuntu 26.04 (WSL2, x86_64, GCC 15.2.0) without GPU/CUDA. The new test fails on `main` (`sem=1`) and passes with this fix, and the existing `postinit_owner_death` test still passes.

I added `test/test_fork_child_exit_cleanup.c` to reproduce the issue.

The test uses `clone(SIGCHLD)` instead of `fork()` because a normal `fork()` runs `pthread_atfork(..., child_reinit_flag)`, which resets `init_status` and causes `exit_handler` to return early. Using `clone(SIGCHLD)` reproduces the path where the child inherits the initialized state without running the atfork handler.

With the original code, the child exits and incorrectly releases the parent's semaphore:

```text id="9b8t1v"
after clone-child exit: child posted parent's lock (sem=1)
exit:1
```

With the fix, the test verifies that after the child exits:

* `sem == 0`, so the parent's lock is still held.
* `owner_pid` is still the parent's PID.
* The parent's PID remains in `procs[]`.

The parent then releases the lock normally, and another process successfully initializes, checking that the lock is not left stuck.

```text id="vgv9j3"
fork-child exit cleanup tests passed
exit:0
```

The test follows the existing GPU-free test setup, so CUDA/NVML hardware is not required.

This fix is limited to inherited cleanup state in child processes that do not run the registered `pthread_atfork` child handler. The normal `fork()` path already resets the initialization state through `child_reinit_flag`.

**AI usage**

I used ChatGPT to help translate my PR description into English and Cursor for assistance while writing the code. I personally reviewed the code, reproduced the issue, ran the tests, and checked the overall behavior and results myself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved process cleanup and identity handling after fork and clone operations.
  * Prevented child processes from releasing a parent’s shared lock or clearing its process slot.
  * Improved reliability when child processes exit or reinitialize after creation.

* **Tests**
  * Added GPU-free regression coverage for fork and clone cleanup behavior.
  * Added automated validation for shared-region ownership and process reinitialization.

### DIFF
--- a/src/multiprocess/multiprocess_memory_limit.c
+++ b/src/multiprocess/multiprocess_memory_limit.c
@@ -858,7 +858,18 @@ void exit_handler() {
         return;
     }
 
-    int32_t my_pid = region_info.pid;
+    /*
+     * Identity is the calling process, not the cached parent pid.  After
+     * clone()/fork paths that skip pthread_atfork, region_info.pid still
+     * belongs to the parent; using it here would drop the parent's slot
+     * and sem_post a lock the parent still holds.
+     */
+    int32_t my_pid = getpid();
+    if (region_info.pid != 0 && region_info.pid != my_pid) {
+        LOG_DEBUG("Skip inherited exit cleanup; cached pid %d != self %d",
+                  region_info.pid, my_pid);
+        return;
+    }
     LOG_MSG("Cleanup on exit for PID %d", my_pid);
 
     // ========================================================================
@@ -913,7 +924,7 @@ void lock_shrreg() {
 
         if (status == 0) {
             // TODO: irregular exit here will hang pending locks
-            region->owner_pid = region_info.pid;
+            region->owner_pid = (size_t)getpid();
             __sync_synchronize();
             SEQ_POINT_MARK(SEQ_UPDATE_OWNER_OK);
             trials = 0;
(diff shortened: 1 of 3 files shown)


maverick-woo:
Reviewed. LGTM, this can be merged. @archlitchi

archlitchi (approved):
/lgtm

Liauuu:
@archlitchi Thank you for the approval. I resolved the merge conflict with the latest main. I’d appreciate it if you could take another look.

archlitchi (approved):
/lgtm

Issue #277, opened 2026-08-14
Run registered GPU-free CTests in CI


lin121291:
## What happened

PR #248 added the GPU-free `postinit_owner_death` regression test and registered it with CTest.

The current `build-src` workflow runs:

```bash
make build-in-docker
```

This builds the test executable, but the Docker build stops after `build.sh` and never runs CTest. CI can therefore pass when the test compiles even if it would fail at runtime.

## Expected behavior

Registered GPU-free CTests should run in CI, and a test failure or an empty test set should fail the build job.

## Validation

I reproduced the current CI build with `nvidia/cuda:13.3.0-cudnn-devel-ubi8` without an NVIDIA driver or GPU.

Results:

- The existing Docker build passed.
- CTest discovered one test: `postinit_owner_death`.
- The test passed in about 0.65 seconds.
- 20 consecutive runs passed.
- The test binary only depends on `librt`, `libpthread`, and `libc`; it has no CUDA or NVML runtime dependency.
- Running the build followed by CTest in the same container passed.
- With CMake 3.26.5, `--no-tests=error` correctly returned a non-zero status for an empty test directory.

## Proposed direction

Run all registered CTests inside the existing build container after a successful build:

```bash
ctest \
  --test-dir /libvgpu/build \
  --output-on-failure \
  --no-tests=error
```

I would avoid filtering for `postinit_owner_death`, so future registered GPU-free tests are covered automatically.

This change would only execute existing tests. It would not modify production behavior or add new test code.

If maintainers agree with this direction, I would like to prepare the PR.

PR #293, opened 2026-08-25
fix: pass the fan index and timestamps to the v2 hooks

iemAnshuman:
## Problem

`nvmlDeviceGetFanSpeed_v2` and `nvmlDeviceGetRetiredPages_v2` in `src/nvml/nvml_entry.c` still use their v1 argument lists, so they drop `fan` and `timestamps`:

```c
nvmlReturn_t nvmlDeviceGetFanSpeed_v2(nvmlDevice_t device, unsigned int fan, unsigned int *speed);
nvmlReturn_t nvmlDeviceGetRetiredPages_v2(nvmlDevice_t device, nvmlPageRetirementCause_t cause,
    unsigned int *pageCount, unsigned long long *addresses, unsigned long long *timestamps);
```

nothing catches it. `driver_sym_t` is `nvmlReturn_t (*)()` with no prototype, and this file never includes `nvml.h`, so the hook is the only declaration around and there is nothing to disagree with it.

The earlier arguments still arrive. The dropped one reaches the driver as whatever its register holds, and the driver writes through it: the callers `speed` pointer in one case, the `timestamps` array in the other.

You won't see it at the default log level, where the code tail calls without touching `%rdx` and the pointer survives by luck. At `LIBCUDA_LOG_LEVEL >= 4` the `LOG_DEBUG` branch calls a cold path that uses `getpid`, `pthread_self` and `basename`, and the hook only restores the registers its signature mentions. `%r8` goes the same way in `nvmlDeviceGetRetiredPages_v2`. It breaks exactly when you turn on debug logging to find out what is wrong.

## Change

Declare the missing argument in both hooks and pass it on. 

I checked all 223 hooks here against `nvml.h` by argument count. One mismatch is left: `nvmlDeviceRemoveGpu` has the three arguments belonging to `nvmlDeviceRemoveGpu_v2`, while `nvml.h` declares `nvmlDeviceRemoveGpu(nvmlPciInfo_t *pciInfo)`. It's harmless, since the extras go to a function reading fewer, so I kept it out of here. Can send it separately.

## Test

`test/test_nvml_v2_signatures.c` is a CTest that needs no GPU and no driver. It builds the production `nvml_entry.c` into the test and gives it a stub `nvml_library_entry` table with the correct v2 signatures, then calls the hooks through the real prototypes and checks the stub got what the caller sent. Pointers are dereferenced only once they match, so a bad build fails cleanly instead of writing somewhere random.

It needs `LIBCUDA_LOG_LEVEL=4` and exits 2 below that. Without debug logging nothing clobbers the register, and the test would pass on the broken hook.

CI compiles this test but never runs it, since `build.sh` just runs `make`. It's a local guard until #277 lands.

On x86_64 AlmaLinux 9.7, gcc 11.5.0, GNU ld 2.35.2, cmake 3.26.5, CUDA 12.2 headers:

1. `bash ./build.sh` exits 0 and `libvgpu.so` links.
2. `ctest` passes 4 of 4.
3. `LIBCUDA_LOG_LEVEL=4 ./test/test_nvml_v2_signatures` passes all 8 checks.

Revert the two hunks and the same test will fail:

```
  driver received fan=1 speed=(nil)
  FAIL speed pointer reaches the driver unchanged
  FAIL timestamps pointer reaches the driver unchanged
```

These hooks only pass telemetry through, so I read gate 2 as not applying and didn't run a GPU cell. cpplint will flag `runtime/int` on the `unsigned long long` arguments; that type comes from the NVML prototype, so I left it.

---
I used deepseek v4 flash to write the code. The logic and the approach are mine, and I ran all the tests myself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fan speed queries now correctly target the requested fan.
  * Retired-page queries now return associated timestamps.
  * Improved forwarding of query parameters and output data for more reliable NVML results.

* **Tests**
  * Added GPU-independent regression coverage for updated fan speed and retired-page queries.
  * Added automated validation of returned values without requiring a physical GPU.
  * Added focused test execution paths with timeouts for improved validation reliability.

### DIFF
--- a/src/nvml/nvml_entry.c
+++ b/src/nvml/nvml_entry.c
@@ -345,10 +345,10 @@ nvmlReturn_t nvmlDeviceGetFanSpeed(nvmlDevice_t device, unsigned int *speed) {
                          speed);
 }
 
-nvmlReturn_t nvmlDeviceGetFanSpeed_v2(nvmlDevice_t device,
+nvmlReturn_t nvmlDeviceGetFanSpeed_v2(nvmlDevice_t device, unsigned int fan,
                                       unsigned int *speed) {
   return NVML_OVERRIDE_CALL(nvml_library_entry, nvmlDeviceGetFanSpeed_v2, device,
-                         speed);
+                         fan, speed);
 }
 
 nvmlReturn_t nvmlDeviceGetFieldValues(nvmlDevice_t device, int valuesCount,
@@ -1212,9 +1212,10 @@ nvmlReturn_t nvmlDeviceGetGridLicensableFeatures_v2(
 nvmlReturn_t nvmlDeviceGetRetiredPages_v2(nvmlDevice_t device,
                                           nvmlPageRetirementCause_t cause,
                                           unsigned int *pageCount,
-                                          unsigned long long *addresses) {
+                                          unsigned long long *addresses,
+                                          unsigned long long *timestamps) {
   return NVML_OVERRIDE_CALL(nvml_library_entry, nvmlDeviceGetRetiredPages_v2,
-                         device, cause, pageCount, addresses);
+                         device, cause, pageCount, addresses, timestamps);
 }
 
 nvmlReturn_t nvmlDeviceResetGpuLockedClocks(nvmlDevice_t device) {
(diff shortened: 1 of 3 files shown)


archlitchi (approved):
/lgtm

PR #295, opened 2026-08-26
chore: add Apache-2.0 LICENSE

mesutoezdil:
HAMi-core has no license file: `gh api repos/Project-HAMi/HAMi-core --jq .license` returns `null`, and one has never existed in git history. This fails OSPS-LE-02.01 and OSPS-LE-03.01 on LFX Insights.

Adds Apache-2.0, byte-identical to `Project-HAMi/HAMi`, so the whole project stays under one license.

Also adds a `NOTICE` file (referenced by section 4 of the Apache License) so the LICENSE is not read as covering third-party code: `src/include/nvml-subset.h` is derived from the NVIDIA Management Library headers and stays under the terms in its own header.

## Summary by CodeRabbit

* **Documentation**
  * Added the Apache License, Version 2.0 terms to the project.
  * Added a notice identifying third-party NVIDIA header licensing and attribution requirements.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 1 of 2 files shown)


maverick-woo:
One concern: `src/include/nvml-subset.h` may still be governed by its original NVIDIA license terms, so Apache-2.0 may not cover that file. Please clarify this or add a third-party notice.

archlitchi (approved):
/lgtm

PR #296, opened 2026-08-26
chore: harden CI workflows and add dependabot config

mesutoezdil:
Neither workflow declared `permissions`, so both jobs ran with a write-capable `GITHUB_TOKEN` (OSPS-AC-04.01), and `actions/checkout` plus `reviewdog/action-cpplint` were pinned to `@master`, a mutable ref, while the hook consistency job still used the deprecated `checkout@v2`. Adds a read-only default with `checks: write` only where reviewdog needs it, pins every action to a commit SHA, and adds `dependabot.yml` so the pins stay current.

Does not fix OSPS-QA-02.01 or OSPS-BR-07.01: both need repository settings, tracked in #299.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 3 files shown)


archlitchi (approved):
/lgtm

PR #297, opened 2026-08-26
docs: add security policy and code of conduct

mesutoezdil:
HAMi-core documents no way to report a vulnerability and has no code of conduct, failing OSPS-VM-02.01 and OSPS-GV-02.01, with the GitHub community profile at 37 percent. Reports point at the HAMi advisory form because private vulnerability reporting is enabled there but not here.

The scope section records that HAMi-core is a cooperative quota mechanism rather than a hard isolation boundary, so quota-only reports are triaged as bugs and cross-tenant reports stay in scope.


archlitchi (approved):
/lgtm

PR #298, opened 2026-08-26
chore: add OpenSSF security-insights.yml

mesutoezdil:
LFX fails OSPS-QA-04.01 with "Insights does not contain a list of repositories" and OSPS-BR-03.02 because no distribution channel is declared anywhere. Adds a spec 2.2.0 `security-insights.yml` naming the two codebases compiled into a HAMi release and recording that HAMi-core ships as `projecthami/hamicore`, carrying an SPDX SBOM, SLSA provenance v1 and a keyless cosign signature.

Merge after #295 and #297, whose files it links to. See the comment below for the correction to my original claim that HAMi-core had no release of its own.

### DIFF
--- a/security-insights.yml
+++ b/security-insights.yml
@@ -1,0 +1,117 @@
+header:
+  schema-version: 2.2.0
+  last-updated: '2026-08-26'
+  last-reviewed: '2026-08-26'
+  url: https://raw.githubusercontent.com/Project-HAMi/HAMi-core/main/security-insights.yml
+  comment: |
+    HAMi-core is one of the codebases that make up the HAMi project. It is
+    consumed as the libvgpu git submodule by Project-HAMi/HAMi, which builds
+    and releases it, so the project section below describes HAMi.
+
+project:
+  name: HAMi
+  homepage: https://project-hami.io
+  administrators:
+    - name: limengxuan
+      affiliation: Dynamia AI
+      social: https://github.com/archlitchi
+      primary: true
+  documentation:
+    quickstart-guide: https://project-hami.io/docs/installation/online-installation
+    detailed-guide: https://project-hami.io/docs
+    code-of-conduct: https://github.com/Project-HAMi/community/blob/main/CODE-OF-CONDUCT.md
+  repositories:
+    - name: HAMi
+      url: https://github.com/Project-HAMi/HAMi
+      comment: |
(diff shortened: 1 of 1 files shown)


mesutoezdil:
Correcting my own PR. I had written that HAMi-core "is not released on its own" and listed only `projecthami/hami` as a distribution point. That was wrong, and it understated what this project already does.

HAMi-core **is** released as its own image. `.github/workflows/call-release-image-hamicore.yaml` in `Project-HAMi/HAMi` builds `docker/Dockerfile.hamicore` from the `libvgpu` submodule and pushes `projecthami/hamicore`, which has 19 tags through `v2.10.0`:

```console
$ curl -s https://hub.docker.com/v2/repositories/projecthami/hamicore/tags | jq '.count'
19
```

Verified against the registry, that image carries more than I credited it with:

```console
$ # in-toto predicate types on the v2.10.0 index
https://spdx.dev/Document
https://slsa.dev/provenance/v1

$ # OCI 1.1 referrers on the same digest
application/vnd.dev.sigstore.bundle.v0.3+json
```

So there is an SPDX SBOM, SLSA provenance v1, and a keyless cosign signature, per platform for `linux/amd64` and `linux/arm64`.

Worth noting for anyone checking this by hand: the cosign signature is stored as an **OCI 1.1 referrer**, not as a legacy `sha256-<digest>.sig` tag. Looking for the `.sig` tag returns 404 and makes it look unsigned. It is not.

The file now records `automated-pipeline: true`, lists `pkg:docker/projecthami/hamicore` as the primary distribution point, declares both attestations with the predicate URIs actually present on the image, and gives the `cosign verify` command. Signature details went into the distribution point comment rather than the `attestations` list, since a sigstore bundle is not an in-toto predicate.

This matters beyond accuracy: OSPS-BR-06.01 and OSPS-QA-02.02 are about signed releases and SBOMs, and the project already satisfies both for this component. Nothing was declaring it.

#297 had the same wrong sentence in `SECURITY.md`; corrected there too.

maverick-woo:
The current Cosign identity regex is too broad and creates a security risk because it trusts signatures from any workflow in `Project-HAMi/HAMi`. Please restrict it to the specific release workflow, `.github/workflows/call-release-image-hamicore.yaml`, and the expected release ref.

archlitchi (approved):
/lgtm

PR #300, opened 2026-08-26
fix: pass cpplint line length through a real action input

mesutoezdil:
reviewdog/action-cpplint has no `args` input, so `args: --linelength=120` has always been dropped and cpplint has been running at its default 80 columns. Over `src` and `test` that is 686 `whitespace/line_length` findings instead of 85.

Moves the setting to `flags`, carrying over the action default `--extensions=h,hpp,c,cpp,cc,cu,hh,ipp` since `flags` replaces that default rather than adding to it.


archlitchi (approved):
/lgtm

PR #301, opened 2026-08-26
chore: pin CUDA base image by digest

mesutoezdil:
`nvidia/cuda:13.3.0-cudnn-devel-ubi8` was pulled by mutable tag in both `dockerfiles/Dockerfile` and the `build-in-docker` target, so rebuilding the same commit can land on different base image contents. `Project-HAMi/HAMi` already pins this exact image by digest in `docker/Dockerfile.hamicore`, the file that builds the released `projecthami/hamicore`.

Uses that same digest, which is what the tag resolves to today, so local and release builds stop drifting apart. #296 adds the `dependabot.yml` that keeps it moving.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,9 @@
 
 current_dir := $(dir $(abspath $(firstword $(MAKEFILE_LIST))))
 
+# Keep in sync with dockerfiles/Dockerfile and Project-HAMi/HAMi's NVIDIA_IMAGE.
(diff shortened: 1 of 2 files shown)


maverick-woo:
Dependabot will update the image in `dockerfiles/Dockerfile`, but it will not update the custom `CUDA_IMAGE` variable in the Makefile. This can make the two image references drift apart. Please use a single source of truth or add a consistency check.

archlitchi (approved):
/lgtm

PR #302, opened 2026-08-26
feat: add CodeQL C/C++ analysis

mesutoezdil:
HAMi runs CodeQL with `language: ["go"]`, so HAMi-core's C is never analysed, despite 16 memory safety fixes in the last 80 commits. Runs in the same CUDA image as `build-in-docker` since `test/CMakeLists.txt` needs nvcc, with `build-mode: manual` to trace the real `build.sh`; no `pull_request` trigger, because fork PRs would sit in `action_required` awaiting approval.

Verified end to end on a fork before opening this ([run 32965369166](https://github.com/mesutoezdil/HAMi-core/actions/runs/32965369166), all steps green), where it found the one issue fixed in #303.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,52 @@
+---
+name: CodeQL
+
+on:
+  push:
+    branches: [main]
+  schedule:
+    - cron: '30 3 * * 1'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


maverick-woo:
The CUDA image digest is now duplicated in this workflow, the Makefile, and the Dockerfile. Dependabot does not update workflow `container.image` references, so these values may drift apart. Please add a consistency check or another update mechanism.

archlitchi (approved):
/lgtm

PR #303, opened 2026-08-26
fix: compute total cuda cores in int64_t

mesutoezdil:
`g_sm_num` and `g_max_thread_per_sm` are `int`, so the product was evaluated in `int` and only widened on assignment to the `int64_t` `g_total_cuda_cores`, while `delta()` in the same file already casts every operand for exactly this reason. No real GPU overflows (H100 8650752, B200 9699328, and `int` math only diverges past 32768 SMs), so this is consistency with `delta()` rather than a live bug.

Found by the CodeQL workflow in #302 as `cpp/integer-multiplication-cast-to-long`.


dongjiang1989 (approved):
Thanks

archlitchi (approved):
/lgtm

PR #304, opened 2026-08-26
feat: account for CUDA context memory

iemAnshuman:
Split out of #251 at @maverick-woo's request. This PR covers context accounting; the fallback lock is #316, and the broker client follows. Refs Project-HAMi/HAMi#1662.

## Problem

HAMi charges primary context memory using a 0/1 activation flag per device. That flag cannot track nested retains or distinguish an intermediate release from the final release. A fork child also inherits bookkeeping for contexts it did not create.

## Change

`context_accounting.c` keeps a retain count and the charged bytes for each device. It charges once and removes the charge after the final successful release. If a shared accounting update fails, it preserves the retain and defers the charge, or keeps the charge for a later removal attempt.

`cuDevicePrimaryCtxReset_v2` removes the resource charge while preserving retained references. CUDA reset does not release retained usage, so a release after reset must still account for those references when another retain has recreated the context.

Each device uses `context_size` from the existing host PID probe. The retain hook does not take a before/after NVML sample: allocation hooks do not share its mutex, so that delta can include unrelated application memory. This keeps main's existing size estimate. Exclusive measurement of a separate size for each device remains outside this PR.

The fork registration and child callback live in `context_fork.c`. `preInit()` calls that registration function. The child clears its private accounting, resets post-init state and keeps the probe estimate. `ctx_activate` is removed because the retain count replaces it.

The allocator and memory mapping hardening from #251 is separate and is not included here.

## Tests

Three tests build production code with no GPU or driver needed at runtime:

1. `context_accounting` covers retain/release bookkeeping, deferred charges, rollback and recovery.
2. `context_accounting_hooks` drives the production retain, release and reset hooks with simulated driver and accounting boundaries. It covers reset followed by retain/release, unrelated allocation during a retain, a refused charge followed by recovery, failed charge removal and driver errors.
3. `context_accounting_fork` calls the production registration function, seeds accounting on two devices and forks. It checks fresh child accounting, unchanged parent references, reset post-init state and another fork in each process.

The Docker build runs these tests with CTest. Its filter also includes `hostpid_fallback_lock` so this PR and #316 use the same command and can merge in either order. [Linux CI](https://github.com/Project-HAMi/HAMi-core/actions/runs/34756983125) passed all three at `7f99005`. Local checks also confirmed that the new regressions reject the corresponding defective implementations. These checks do not validate GPU behavior or a complete HAMi workload.

```sh
cmake --build build --target test_context_accounting test_context_accounting_hooks test_context_accounting_fork
ctest --test-dir build --output-on-failure -R context_accounting
```

---
I used AI assistance (deepseek v4 flash) for the original split and Codex for these fixes. The accounting design comes from #251.

### DIFF
--- a/src/cuda/context.c
+++ b/src/cuda/context.c
@@ -210,9 +210,9 @@ CUresult cuDevicePrimaryCtxRelease_v2( CUdevice dev ){
                                   cuDevicePrimaryCtxRelease_v2, dev);
     }
     pthread_mutex_lock(&context_device_locks[dev]);
-    pthread_mutex_lock(&context_accounting_lock);
     CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,cuDevicePrimaryCtxRelease_v2,dev);
     if (res == CUDA_SUCCESS) {
+        pthread_mutex_lock(&context_accounting_lock);
         if (primary_context_record_release(&context_accounting[dev],
                                            &bytes_to_remove) != 0) {
             LOG_WARN("Unbalanced primary context release on device %d", dev);
@@ -224,8 +224,8 @@ CUresult cuDevicePrimaryCtxRelease_v2( CUdevice dev ){
             }
         }
         ctx_activate[dev] = (int)context_accounting[dev].retain_count;
+        pthread_mutex_unlock(&context_accounting_lock);
     }
-    pthread_mutex_unlock(&context_accounting_lock);
     pthread_mutex_unlock(&context_device_locks[dev]);
     return res;
 }
--- a/src/cuda/context.c
+++ b/src/cuda/context.c
@@ -1,3 +1,5 @@
+#include <errno.h>
+
 #include "include/libcuda_hook.h"
 #include "include/libvgpu.h"
 #include "cuda/context_accounting.h"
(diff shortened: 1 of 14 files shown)


Priyanshi507:
Built `feat-context-accounting` and ran the test suite:

$ ctest --test-dir build -R context_accounting --output-on-failure
100% tests passed, 0 tests failed out of 1

Also read through the locking for cross-process implications, since this
touches the #1662 path: `context_accounting_lock` releases before the
driver call/NVML measurement, so it never holds the expensive path.
`context_device_locks[dev]` does span the driver call, but it's a plain
`static pthread_mutex_t` (no `PTHREAD_PROCESS_SHARED`), confirmed by the
single `pthread_atfork` registration resetting it per-process on fork —
so it serializes threads within one process, not across processes.
Measurement is scoped to the caller's own PID via
`get_used_gpu_memory_by_pid`, so a sibling process shouldn't leak into
the delta.

One minor gap: `test_forked_child_resets_private_accounting` calls
`primary_context_accounting_reset()` directly inside a manually-written
`fork()`, so it verifies the reset logic itself but doesn't exercise the
actual `pthread_atfork` → `childReinitPostInit` →
`context_accounting_fork_child` wiring that runs it in practice.

iemAnshuman:
@Priyanshi507 thanks for the linux run. yeah I am already working on that one and it is part of why this is still a draft.

Priyanshi507:
> @Priyanshi507 thanks for the linux run. yeah I am already working on that one and it is part of why this is still a draft.

Makes sense — thanks for confirming. I'll leave that one to you and won't duplicate effort digging into it further.

src/cuda/context.c
```diff
@@ -11,16 +52,84 @@ CUresult cuDevicePrimaryCtxGetState( CUdevice dev, unsigned int* flags, int* act
...
+    }
+
+    pthread_mutex_lock(&context_accounting_lock);
+    /* Allocation hooks do not share this lock, so a before/after NVML delta
+     * can include unrelated memory. Use the established probe charge on each
+     * device until an exclusive measurement is available. */
+    charge = context_size;
+    errno = 0;
+    int record_result =
+        (pidfound == 1)
```
mesutoezdil: the ENODATA retry lands on record_retain either way, so pidfound only changes a log line here. is it doing more than that?

iemAnshuman: you are right that it only changed a log line. it was left over from #251, where ENODATA used to fail the retain so I dropped the pidfound branch and the accounted variant in fdac406 and 0c70a7e

iemAnshuman:
the pidfound check is gone. when a charge gets refused or a removal fails, the hook now just sets charged_bytes directly, so I removed the rollback and restore helpers too. context_fork.c is folded into context.c now, and I took the ctest filter out of the Makefile since main already runs all the tests.

iemAnshuman:
on the test side I dropped a few checks that could never fail, and added cases for a failed removal on release, retaining again after the last release, and devices outside the accounting range.

mesutoezdil:
point 4

iemAnshuman:
> This is being closed because it does not comply with the contribution guidelines. point 4

those two comments were improved using ai. that's 4th point, sorry for that. not doing that anymore.  is there anything you want me to change before you look at this again?

PR #309, opened 2026-08-28
fix: avoid SIGSEGV when set_task_pid() is bypassed due to lock failure (postInit)

Bk88416:
## Why

When `lock_postInit()` fails in `postInit()`, `set_task_pid()` is skipped, which bypasses `nvmlInit()`. This prevents `nvml_preInit()` → `load_nvml_libraries()` from executing, leaving `nvml_library_entry[]` uninitialized.

Later, `init_utilization_watcher()` → `nvmlDeviceGetCount()` dereferences the null function pointer and causes a SIGSEGV.

## Fix

1. Call `nvmlInit()` unconditionally in `postInit()` to ensure `load_nvml_libraries()` always runs.
2. Add null-check protection in `*_OVERRIDE_CALL()` macro for robustness.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when CUDA or NVML driver functions are unavailable, returning clear error results instead of attempting invalid calls.
  * Enhanced stability when requested driver symbols cannot be resolved.
  * Ensured NVML initialization occurs during device setup for more reliable GPU monitoring and management.

### DIFF
--- a/src/include/libcuda_hook.h
+++ b/src/include/libcuda_hook.h
@@ -36,6 +36,7 @@ typedef CUresult (*cuda_sym_t)();
     cuda_sym_t _entry = (cuda_sym_t)CUDA_FIND_ENTRY(table, sym);               \
     if (_entry == NULL) {                                                      \
       LOG_ERROR("Hijack failed: %s is NULL", #sym);                            \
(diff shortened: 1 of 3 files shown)


mesutoezdil:

PR #311, opened 2026-08-28
fix(allocator): use tracked allocation device instead of current context in async free accounting (#310)

### LINKED ISSUE #310: remove_chunk_async decrements the wrong device's tracked memory usage when the current context's device differs from the allocation's device
## Background

`src/allocator/allocator.c`, `remove_chunk_async()` (lines 240–259), reached via `free_raw_async()` ← the `cuMemFreeAsync` hook in `src/cuda/memory.c:637`.

## Bug

```c
int remove_chunk_async(
    allocated_list *a_list, CUdeviceptr dptr, CUstream hStream) {
    size_t t_size;
    allocated_list_entry *val;
    for (val = a_list->head; val != NULL; val = val->next) {
        if (val->entry->address == dptr) {
            t_size=val->entry->length;
            CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemFreeAsync,dptr,hStream);
            LIST_REMOVE(a_list,val);
            a_list->limit-=t_size;
            CUdevice dev;
            cuCtxGetDevice(&dev);                      // <-- BUG
            rm_gpu_device_memory_usage(getpid(),dev,t_size,2);
            return 0;
        }
    }
```

When an async allocation is freed, the function determines which device's usage counter to decrement via `cuCtxGetDevice(&dev)` — the device of the **calling thread's currently-bound context** — rather than `val->entry->dev`, the device the chunk was actually allocated on and accounted against (populated at allocation time by `INIT_ALLOCATED_LIST_ENTRY`, see `allocator.h`).

Its sibling function, the synchronous `remove_chunk()` a few lines above in the same file, gets this right:

```c
t_dev = val->entry->dev;
...
rm_gpu_device_memory_usage(getpid(), t_dev, t_size, 2);
```

`remove_chunk_async` diverges from that pattern and substitutes the current context's device instead.

## Why this is reachable in practice

`cuMemFreeAsync(dptr, hStream)` takes an explicit stream, not an implicit "current device." CUDA's own API contract does not require the calling thread's current context to match the device the stream (or the memory) belongs to — stream handles can be created on one device and freed from code paths where a different device happens to be current (multi-GPU pipelines, worker-thread pools that round-robin `cuCtxSetCurrent`, async cleanup callbacks). HAMi-core does nothing to prevent this: `cuCtxSetCurrent` is hooked as a pure passthrough with no tracking, and `cuMemFreeAsync` is hooked as a direct passthrough to `free_raw_async` with no device pinning or validation before it reaches `remove_chunk_async`. The README confirms multi-GPU-per-container (`CUDA_DEVICE_MEMORY_LIMIT` "for all devices") is a supported configuration, not an edge case outside this library's scope.

## Impact

`rm_gpu_device_memory_usage(pid, dev, size, 2)` decrements `used[dev]` in the per-process, per-device shared-memory usage table (`multiprocess_memory_limit.c`) that `get_gpu_memory_usage()`/`oom_check()` read to enforce the vGPU memory limit. A device mismatch here corrupts accounting for **two** devices at once:

- The real allocation device (`val->entry->dev`) never gets decremented → its tracked usage stays permanently inflated even though the memory was actually freed, eventually causing spurious OOM rejections for legitimate future allocations on that device.
- The wrong device (`dev`, from `cuCtxGetDevice`) gets decremented for memory it never held → its tracked usage under-counts real consumption, which can let a process exceed its actual configured memory limit on that device undetected — defeating the enforcement this library exists to provide.

Over a long-running multi-GPU workload, repeated mismatches compound: one device's phantom usage climbs indefinitely, another's drifts arbitrarily below reality.

## Suggested fix

Capture `val->entry->dev` into a local before `LIST_REMOVE` frees `val->entry` (mirroring `remove_chunk`), and pass that captured device to `rm_gpu_device_memory_usage` instead of calling `cuCtxGetDevice`:

```c
t_size = val->entry->length;
CUdevice t_dev = val->entry->dev;   /* capture before LIST_REMOVE frees entry */
CUDA_OVERRIDE_CALL(cuda_library_entry, cuMemFreeAsync, dptr, hStream);
LIST_REMOVE(a_list, val);
a_list->limit -= t_size;
rm_gpu_device_memory_usage(getpid(), t_dev, t_size, 2);
```

## Verification notes

- Confirmed against current `main` (not from memory): `git log -- src/allocator/allocator.c` shows this pattern present since the function was introduced, with no later fix touching it.
- Not a duplicate of #306/#307 — that issue covers leaked allocations on post-allocation driver failure (the allocate path); this is a distinct bug in the free path with a distinct root cause.
- No existing test covers this: `test/test_alloc_pool_async.c` exercises `cuMemFreeAsync` but only on a single device, so current-context and allocation-device always coincide there and the existing suite can't catch this.

AyushSrivastava1818:
## Fixes
Closes #310

## Background
In `src/allocator/allocator.c`, `remove_chunk_async()` (used by the `cuMemFreeAsync` hook via `free_raw_async()`) determines which device's usage counter to decrement by calling `cuCtxGetDevice(&dev)` — the device of the calling thread's *currently-bound context* — instead of using `val->entry->dev`, the device the chunk was actually allocated on and accounted against.

`cuMemFreeAsync(dptr, hStream)` takes an explicit stream, not an implicit "current device," and CUDA does not require the calling thread's current context to match the device the stream or memory belongs to. In any multi-GPU process (worker threads round-robining `cuCtxSetCurrent`, pipelines that allocate on one device and free from a code path where another device happens to be current), this mismatch is fully reachable — nothing in HAMi-core's hook chain pins or validates the device before the call reaches `remove_chunk_async`.

`rm_gpu_device_memory_usage()` decrements `used[dev]` in the per-process, per-device shared-memory usage table that `get_gpu_memory_usage()`/`oom_check()` read to enforce the vGPU memory limit — the core mechanism this library exists to provide. A device mismatch here corrupts accounting for two devices at once: the real allocation device is never decremented (usage stays permanently inflated, eventually causing spurious OOM rejections), while the wrong device is decremented for memory it never held (usage under-counts real consumption, letting a process exceed its actual limit undetected).

The sibling synchronous function, `remove_chunk()`, already does this correctly via `t_dev = val->entry->dev`. `remove_chunk_async` diverged from that pattern.

Full details and impact analysis are in #310.

## Changes
- `src/allocator/allocator.c`: capture `t_dev = val->entry->dev` before `LIST_REMOVE` frees the entry, and pass that to `rm_gpu_device_memory_usage` instead of `cuCtxGetDevice()`. Removed the now-dead `CUdevice dev; cuCtxGetDevice(&dev);` lines. 3-line net diff.
- `test/test_alloc_async_free_wrong_device.c`: new regression test. Allocates on device 0, switches current context to device 1, frees via device 0's stream through the real hooked `cuMemFreeAsync` — the exact mismatch — then reads `get_gpu_memory_usage()` (linking `multiprocess_memory_limit.c` directly into the test binary, same pattern as `test_postinit_owner_death`) to assert device 0's usage returns to baseline, device 1's stays untouched, and a subsequent clean allocation on device 1 isn't affected by any lingering corruption. Skips gracefully via CTest's `SKIP_RETURN_CODE` on environments with fewer than 2 GPUs.
- `test/CMakeLists.txt`: registered the new test target, reusing the plain `LD_PRELOAD=libvgpu.so` setup — no fault-injection shim needed since this test exercises a genuine two-device code path rather than forcing a driver call to fail.

## Verification
- `cpplint --linelength=120 src/allocator/allocator.c`: no new warnings (diffed precisely against the pre-fix file; same 5 pre-existing warnings, unrelated, shifted by line number only — the new/changed lines are clean).
- `python3 hack/check_cuda_hook_consistency.py`: PASS.
- Traced the test's expected accounting deltas by hand against `add_gpu_device_memory_usage`/`rm_gpu_device_memory_usage` to confirm the assertions are deterministic and not dependent on driver-specific quirks.
- Could not compile or run anything in this sandbox — no CUDA toolchain, no GPU, and no `cuda.h` available (`gcc -fsyntax-only` fails immediately on the missing header). This PR has **not** been built or executed against a real CUDA driver or real multi-GPU hardware. Requesting CI/maintainer verification before merge, same as #307.

## AI disclosure
This issue, fix, and PR description were produced with substantial AI assistance (Claude): the bug was found via AI-driven source review of `allocator.c`, the fix and the new regression test were written by an AI coding agent, and this PR body was drafted by AI. All of it has been reviewed and is being submitted by me as the human author of record — I take responsibility for its correctness. Per the Verification section above, this fix is unvalidated against real CUDA hardware due to the sandboxed environment's lack of a CUDA toolchain/GPU; I'd appreciate maintainer/CI scrutiny on the multi-GPU test behavior in particular before merge.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected asynchronous memory cleanup so usage tracking is updated for the device where memory was allocated, even when freed from another device context.
  - Prevented inaccurate device memory accounting when allocations are released from a different active device.

- **Tests**
  - Added regression coverage for freeing allocations across different device contexts.
  - Added safeguards for environments with fewer than two available CUDA devices, where the test is skipped.

### DIFF
--- a/src/allocator/allocator.c
+++ b/src/allocator/allocator.c
@@ -244,12 +244,11 @@ int remove_chunk_async(
     for (val = a_list->head; val != NULL; val = val->next) {
         if (val->entry->address == dptr) {
             t_size=val->entry->length;
+            CUdevice t_dev = val->entry->dev;  /* capture before LIST_REMOVE frees entry */
             CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemFreeAsync,dptr,hStream);
             LIST_REMOVE(a_list,val);
             a_list->limit-=t_size;
-            CUdevice dev;
-            cuCtxGetDevice(&dev);
-            rm_gpu_device_memory_usage(getpid(),dev,t_size,2);
+            rm_gpu_device_memory_usage(getpid(),t_dev,t_size,2);
             return 0;
         }
     }
--- a/src/allocator/allocator.c
+++ b/src/allocator/allocator.c
@@ -248,7 +248,7 @@ int remove_chunk_async(
             CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemFreeAsync,dptr,hStream);
             LIST_REMOVE(a_list,val);
             a_list->limit-=t_size;
-            rm_gpu_device_memory_usage(getpid(),t_dev,t_size,2);
+            rm_gpu_device_memory_usage(getpid(), t_dev, t_size, 2);
             return 0;
         }
     }
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -32,6 +32,24 @@ foreach(TEST_SCRIPT ${TEST_SCRIPTS})
             -ffunction-sections -fdata-sections)
         set_target_properties(${TEST_TARGET_NAME} PROPERTIES
             LINK_FLAGS "-Wl,--no-export-dynamic,--gc-sections")
+    elseif (TEST_TARGET_NAME STREQUAL "test_alloc_async_free_wrong_device")
+        # Needs get_gpu_memory_usage() from multiprocess_memory_limit.c linked
+        # directly into the test binary (same reasoning as
+        # test_postinit_owner_death above), on top of the normal CUDA driver
+        # link every allocation test needs. -ffunction-sections/--gc-sections
+        # let the linker drop the parts of multiprocess_memory_limit.c this
+        # test never calls, instead of requiring their own transitive deps
+        # (e.g. cuda_to_nvml_map from multiprocess_utilization_watcher.c).
+        add_executable(${TEST_TARGET_NAME}
+            ${TEST_SCRIPT}
+            ${CMAKE_CURRENT_SOURCE_DIR}/../src/multiprocess/multiprocess_memory_limit.c
+            ${CMAKE_CURRENT_SOURCE_DIR}/../src/log_utils.c)
+        target_compile_definitions(${TEST_TARGET_NAME} PRIVATE
+            _GNU_SOURCE)
+        target_compile_options(${TEST_TARGET_NAME} PRIVATE
+            -ffunction-sections -fdata-sections)
+        set_target_properties(${TEST_TARGET_NAME} PROPERTIES
+            LINK_FLAGS "-Wl,--no-export-dynamic,--gc-sections")
     elseif (TEST_TARGET_NAME STREQUAL "test_dlsym_rtld_next")
         add_executable(${TEST_TARGET_NAME} ${TEST_SCRIPT})
     elseif (TEST_SCRIPT MATCHES ".*cu")
@@ -46,6 +64,8 @@ foreach(TEST_SCRIPT ${TEST_SCRIPTS})
     elseif (TEST_TARGET_NAME STREQUAL "test_dlsym_rtld_next")
         target_link_libraries(${TEST_TARGET_NAME} -ldl)
     else()
+        # test_alloc_async_free_wrong_device also lands here: it needs the
+        # same CUDA driver link as every other allocation test.
         target_link_libraries(${TEST_TARGET_NAME} -lrt -lpthread
             -lnvidia-ml -lcuda -lcudart -L${CUDA_HOME}/lib64)
     endif()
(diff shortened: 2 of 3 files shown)


AyushSrivastava1818:
@archlitchi have a look Sir...if theres no issue I would resolve the conflicts then get it merged

src/allocator/allocator.c
```diff
@@ -244,12 +244,11 @@ int remove_chunk_async(
     for (val = a_list->head; val != NULL; val = val->next) {
         if (val->entry->address == dptr) {
             t_size=val->entry->length;
+            CUdevice t_dev = val->entry->dev;  /* capture before LIST_REMOVE frees entry */
```
mesutoezdil, line 247: #290 already rewrites remove_chunk_async the same way. did you two settle which one keeps it?

AyushSrivastava1818: not discussed yet will do

Eshiv-Pandey: @AyushSrivastava1818 yoo hi, #290 has had this since ee875ef I'd say it stays there. your 2-device test is worth landing on top though, once rebased. works?

AyushSrivastava1818: @Eshiv-Pandey Yes you can go on i will do the required changes once yours gets merged!

mesutoezdil: ee875ef landed this on main. only the test is new now.

mesutoezdil: drop this hunk, keep the test. main has no coverage for it.

PR #313, opened 2026-08-28
fix(hooks): guard driver-call out-params behind a success check (#312)

### LINKED ISSUE #312: cuMemPoolGetAttribute hook dereferences the caller's value pointer as a fixed 8-byte long regardless of call success or the attribute's real type width
## Background

`src/cuda/memory.c`, `cuMemPoolGetAttribute()` (lines 668-672), a wired hook (`hook.c:150`, `libcuda_hook.h:184`) intercepting every application call to the real `cuMemPoolGetAttribute`.

## Bug

```c
CUresult cuMemPoolGetAttribute(CUmemoryPool pool, CUmemPool_attribute attr, void *value) {
    CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemPoolGetAttribute,pool,attr,value);
    LOG_INFO("cuMemPoolGetAttribute %d %ld",attr,*(long *)value);
    return res;
}
```

Two independent problems in one line:

1. **Read-before-success-check.** `*(long *)value` is dereferenced unconditionally, even when `res != CUDA_SUCCESS`. On failure, the driver has no obligation to have written anything to `*value`, so this reads whatever was already in the caller's buffer — uninitialized stack/heap memory if the caller declared it without a default value (a very common pattern: `size_t v; cuMemPoolGetAttribute(pool, attr, &v);`).

2. **Fixed-width cast ignores the attribute's real type.** `void *value`'s actual pointee type depends on which `CUmemPool_attribute` was passed — this is inherent to the CUDA driver API's design, not a HAMi-core choice. Per NVIDIA's documented `CUmemPool_attribute` semantics, `CU_MEMPOOL_ATTR_REUSE_FOLLOW_EVENT_DEPENDENCIES`, `CU_MEMPOOL_ATTR_REUSE_ALLOW_OPPORTUNISTIC`, and `CU_MEMPOOL_ATTR_REUSE_ALLOW_INTERNAL_DEPENDENCIES` take an `int*` (4 bytes); the remaining attributes (`CU_MEMPOOL_ATTR_RELEASE_THRESHOLD`, `CU_MEMPOOL_ATTR_RESERVED_MEM_CURRENT`, `CU_MEMPOOL_ATTR_RESERVED_MEM_HIGH`, `CU_MEMPOOL_ATTR_USED_MEM_CURRENT`, `CU_MEMPOOL_ATTR_USED_MEM_HIGH`) take a `cuuint64_t*` (8 bytes). A caller correctly following the API contract for one of the `int`-typed attributes passes a 4-byte buffer. This wrapper's `*(long *)value` then reads 8 bytes — running 4 bytes past the caller's actual variable **even when the call succeeds**, not only on failure.

I confirmed the `CUmemPool_attribute` enum itself (names/ordering, including newer additions like `CU_MEMPOOL_ATTR_ALLOCATION_TYPE` through `CU_MEMPOOL_ATTR_HW_DECOMPRESS_ENABLED`) against `cydriver.pxd` from the `cuda-bindings` PyPI package, which is Cython-generated directly from NVIDIA's real `cuda.h`. That artifact confirms the enum and the `void *value` signature but does not itself carry a per-attribute width table — the int-vs-`cuuint64_t` split above is long-published, stable NVIDIA API documentation I'm recalling, not something re-derived from a source file in this environment. Worth a maintainer double-check against a local CUDA toolkit's `cuda.h` doc comments before relying on it for the fix.

## Why this matters despite being a log statement

- Reading 4 bytes past a stack-allocated `int` is undefined behavior: silently harmless in the overwhelmingly common case, but a real bug under AddressSanitizer/stack-protector-strict builds, and a (minor) information disclosure of adjacent stack/heap contents into the log.
- This is the same defect class CodeRabbit already flagged during review of PR #307/#310 — but only in the *test file* that modeled a call to this exact production function ("initialize `attr_val` before the invalid-attribute call... the preload wrapper unconditionally reads `*(long *)value`"). The production function itself was never fixed.

## Scope / honest severity assessment

This is **not** default-on. `LOG_INFO` (see `src/log_utils.h`) is gated behind `g_log_level >= 3`; the default (`src/log_utils.c:8`, `g_log_level = 2`) does not reach this line. It only fires when a user sets `LIBCUDA_LOG_LEVEL=3` (INFO) or higher — a realistic thing to do while specifically debugging memory-pool behavior, but not the out-of-the-box configuration. Filing this as a low/moderate-severity correctness bug, not a security-critical one.

## Suggested fix

Don't assume a fixed width or dereference on failure:

```c
CUresult cuMemPoolGetAttribute(CUmemoryPool pool, CUmemPool_attribute attr, void *value) {
    CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemPoolGetAttribute,pool,attr,value);
    if (res == CUDA_SUCCESS) {
        LOG_INFO("cuMemPoolGetAttribute attr=%d", attr);
    } else {
        LOG_INFO("cuMemPoolGetAttribute attr=%d failed res=%d", attr, res);
    }
    return res;
}
```
(Dropping the value from the log entirely is the simplest safe fix, since correctly reading it back requires switching on `attr` to know whether to print an `int` or a `cuuint64_t`. If the value is worth keeping in the log, that switch should be added explicitly rather than guessed via a blind cast.)

## Verification notes

- Confirmed against current `main`: no later commit touches this line.
- Not a duplicate of #306/#307/#310 — those cover the allocator's tracking-entry lifecycle; this is an unrelated hook with no interaction with `device_allocasync`/`device_overallocated`.
- Scanned all of `src/cuda/memory.c` (846 lines) — this pattern is isolated to this one function, not systemic across the file.
- Could not do a live duplicate-check against the open issue list — GitHub's REST API is rate-limited unauthenticated from this environment. Worth a manual search before triage in case something already covers it.

AyushSrivastava1818:
Fixes #312.

## Background

#312 was filed against `src/cuda/memory.c:668-672`, `cuMemPoolGetAttribute()`, and closed the same day with no linked commit, PR, or comment explaining the close (the only timeline event before "closed" is `/assign`). The bug was still present on `main` at `de6ce39`. I reopened the issue with an explanatory comment before starting this PR.

## Original defect

```c
CUresult cuMemPoolGetAttribute(CUmemoryPool pool, CUmemPool_attribute attr, void *value) {
    CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemPoolGetAttribute,pool,attr,value);
    LOG_INFO("cuMemPoolGetAttribute %d %ld",attr,*(long *)value);
    return res;
}
```

Two independent problems in one line:
1. **Read-before-success-check.** `*(long *)value` is dereferenced unconditionally, even when `res != CUDA_SUCCESS`. On failure the driver has no obligation to have written `*value`, so this reads whatever garbage was already on the caller's stack/heap.
2. **Fixed-width cast ignores the attribute's real type.** `value`'s pointee type depends on which `CUmemPool_attribute` was passed (some are `int*`, some are `cuuint64_t*` per the CUDA driver API). The fixed `(long *)` cast is wrong for the narrower ones even on a successful call.

**On the width table:** the original issue proposed hardcoding a per-attribute int-vs-`cuuint64_t` width table. I did not add it. No CUDA header (`cuda.h`, `cydriver.pxd`, or otherwise) exists anywhere in this repo or its dependency chain to verify that table against -- the only local `cuda.h` is a hand-written, explicitly-non-authoritative `-fsyntax-only` sanity stub from an unrelated prior session. Guessing a width table from memory and shipping it felt like the wrong tradeoff for a log line. Instead this drops the value from the log entirely, which is the fallback the original issue itself proposed:

```c
CUresult cuMemPoolGetAttribute(CUmemoryPool pool, CUmemPool_attribute attr, void *value) {
    CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemPoolGetAttribute,pool,attr,value);
    if (res == CUDA_SUCCESS) {
        LOG_INFO("cuMemPoolGetAttribute attr=%d", attr);
    } else {
        LOG_INFO("cuMemPoolGetAttribute attr=%d failed res=%d", attr, res);
    }
    return res;
}
```

If a maintainer has a real CUDA toolkit handy and wants the value back in the log for the successful case, that needs an explicit `switch` on `attr` to pick the right width -- I'd rather that be added deliberately by someone who can verify it than guessed here.

## Same defect class, three more places (in scope for this PR)

These have problem 1 (unconditional out-param read before checking `res`) but not problem 2 -- their out-param types are fixed by the function signature, not attribute-dependent, so there's no width question:

- **`cuMemGetInfo_v2`** (`src/cuda/memory.c`): `CUDA_OVERRIDE_CALL`'s return value wasn't even captured, and `free`/`total` were logged (twice, in each branch) regardless of whether the driver call actually wrote them. One branch also went on to *compute* with the unwritten values. Now captures `res` and returns early on failure, before any read.
- **`cuMemAddressReserve`** (`src/cuda/memory.c`): `res` was captured but never checked before logging `*ptr`.
- **`cuDeviceGetCount`** (`src/cuda/device.c`): `count` was logged unconditionally (`LOG_DEBUG`, gated `>= 4`).

All four now follow the res-check-first idiom already used correctly elsewhere in this file (`cuDriverGetVersion`, `memory.c`'s `cuMemAlloc_v2`) rather than inventing a new pattern. No other changes to these three functions -- I did not refactor anything beyond the guard.

I did **not** touch two other `*(cast *)` log-deref sites found during the sweep because they're already correctly guarded and aren't part of this defect class: `cuDriverGetVersion` (already checks `res==CUDA_SUCCESS` first) and `cuMemAlloc_v2` (already returns early on failure).

## Tests

The existing PR #307 test infrastructure (`fault_inject_driver.c` / `test_alloc_async_failure_leak.c`) does not cover any of these four wrappers -- it fakes the real driver's `cuDeviceGetMemPool`/`cuMemPoolGetAttribute` only to test `src/allocator/allocator.c`'s bookkeeping, never HAMi-core's own hook in `memory.c`. So it gives no coverage here.

Added, modeled directly on that same precedent:
- `test/fault_inject_driver_312.c` -- an `LD_PRELOAD` companion that interposes `dlopen("libcuda.so.1")` to force one-shot failures out of `cuMemPoolGetAttribute`, `cuMemGetInfo_v2`, `cuMemAddressReserve`, and `cuDeviceGetCount`, forwarding everything else to the real driver.
- `test/test_312_failure_path_log_guard.c` -- poisons each out-param buffer with a distinctive canary value, forces the underlying call to fail, captures HAMi-core's log output (`LOG_INFO`/`LOG_DEBUG` write to `stderr` in the default non-`FILEDEBUG` build), and asserts the canary never appears in it. Also has a `cuMemPoolGetAttribute` success-path check confirming the log still fires and never contains a raw value. Registered in `test/CMakeLists.txt` the same way PR #307 registered its fault-injection companion.

I did not use ASan/UBSan for this: ASan doesn't flag an in-bounds read of unpoisoned-but-uninitialized stack memory (there's no bounds violation), and MemorySanitizer -- the tool that would catch that -- can't be used against a closed-source `libcuda.so.1`. The canary-in-the-log approach directly observes the actual defect (does the wrapper leak an unwritten buffer into its log) rather than relying on a sanitizer that can't see it.

**These tests need a real GPU to build and run, and are not exercised in CI** -- same as every other test in this directory (`.github/workflows/build-src.yml` only builds via `make build-in-docker`; there's no `ctest`/GPU runner step, which PR #307's own test comments already note explicitly).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA error handling so driver failures are correctly propagated.
  * Prevented failure logs from displaying uninitialized or invalid output values.
  * Ensured memory and device operation logs accurately reflect success or failure.

* **Tests**
  * Added regression coverage for CUDA failure paths, error propagation, and safe diagnostic logging.

### DIFF
--- a/src/cuda/device.c
+++ b/src/cuda/device.c
@@ -22,9 +22,9 @@ CUresult cuDeviceGetCount( int* count ) {
     LOG_DEBUG("into cuDeviceGetCount");
     CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,cuDeviceGetCount,count);
     if (res == CUDA_SUCCESS) {
-        LOG_DEBUG("cuDeviceGetCount res=%d count=%d",res,*count);
+        LOG_DEBUG("cuDeviceGetCount res=%d count=%d", res, *count);
     } else {
-        LOG_DEBUG("cuDeviceGetCount res=%d",res);
+        LOG_DEBUG("cuDeviceGetCount res=%d", res);
     }
     return res;
 }
--- a/src/cuda/device.c
+++ b/src/cuda/device.c
@@ -21,7 +21,11 @@ CUresult cuDeviceGet(CUdevice *device,int ordinal){
 CUresult cuDeviceGetCount( int* count ) {
     LOG_DEBUG("into cuDeviceGetCount");
     CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,cuDeviceGetCount,count);
-    LOG_DEBUG("cuDeviceGetCount res=%d count=%d",res,*count);
+    if (res == CUDA_SUCCESS) {
+        LOG_DEBUG("cuDeviceGetCount res=%d count=%d",res,*count);
+    } else {
+        LOG_DEBUG("cuDeviceGetCount res=%d",res);
+    }
     return res;
 }
 
--- a/src/cuda/memory.c
+++ b/src/cuda/memory.c
@@ -490,14 +490,21 @@ CUresult cuMemGetInfo_v2(size_t* free, size_t* total) {
     CHECK_DRV_API(cuCtxGetDevice(&dev));
     size_t usage = get_current_device_memory_usage(cuda_to_nvml_map(dev));
     size_t limit = get_current_device_memory_limit(cuda_to_nvml_map(dev));
+    CUresult res;
     if (limit == 0) {
-        CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemGetInfo_v2, free, total);
+        res = CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemGetInfo_v2, free, total);
+        if (res != CUDA_SUCCESS) {
+            return res;
+        }
         LOG_INFO("orig free=%ld total=%ld", *free, *total);
         *free = *total - usage;
         LOG_INFO("after free=%ld total=%ld", *free, *total);
         return CUDA_SUCCESS;
     } else {
-        CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemGetInfo_v2, free, total);
+        res = CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemGetInfo_v2, free, total);
+        if (res != CUDA_SUCCESS) {
+            return res;
+        }
         LOG_INFO("orig free=%ld total=%ld limit=%ld usage=%ld",
             *free, *total, limit, usage);
         // Ensure total memory does not exceed the physical or imposed limit.
@@ -583,7 +590,9 @@ CUresult cuMemAddressReserve(CUdeviceptr* ptr, size_t size,
     size_t alignment, CUdeviceptr addr, unsigned long long flags ) {
     CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,
         cuMemAddressReserve, ptr, size, alignment, addr, flags);
-    LOG_INFO("cuMemAddressReserve:%lx %llx", size, *ptr);
+    if (res == CUDA_SUCCESS) {
+        LOG_INFO("cuMemAddressReserve:%lx %llx", size, *ptr);
+    }
     return res;
 }
 
@@ -667,7 +676,11 @@ CUresult cuMemPoolSetAttribute(CUmemoryPool pool, CUmemPool_attribute attr, void
 
 CUresult cuMemPoolGetAttribute(CUmemoryPool pool, CUmemPool_attribute attr, void *value) {
     CUresult res = CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemPoolGetAttribute,pool,attr,value);
-    LOG_INFO("cuMemPoolGetAttribute %d %ld",attr,*(long *)value);
+    if (res == CUDA_SUCCESS) {
+        LOG_INFO("cuMemPoolGetAttribute attr=%d", attr);
+    } else {
+        LOG_INFO("cuMemPoolGetAttribute attr=%d failed res=%d", attr, res);
+    }
     return res;
 }
 
--- a/src/cuda/memory.c
+++ b/src/cuda/memory.c
@@ -492,7 +492,7 @@ CUresult cuMemGetInfo_v2(size_t* free, size_t* total) {
     size_t limit = get_current_device_memory_limit(cuda_to_nvml_map(dev));
     CUresult res;
     if (limit == 0) {
-        res = CUDA_OVERRIDE_CALL(cuda_library_entry,cuMemGetInfo_v2, free, total);
+        res = CUDA_OVERRIDE_CALL(cuda_library_entry, cuMemGetInfo_v2, free, total);
         if (res != CUDA_SUCCESS) {
             return res;
         }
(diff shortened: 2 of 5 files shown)


mesutoezdil:

AyushSrivastava1818:
@mesutoezdil so sorryy sir for the inconvenience caused..may I know what was the reason of this being closed Sir?

PR #316, opened 2026-09-03
feat: add a node wide host pid fallback lock

iemAnshuman:
Split out of #251 at @maverick-woo's request. Just the lock here. Broker client and the integration follow.

## Problem

`postInit` serializes host PID discovery with `lock_postinit()`, a record lock on the cache file. That only covers processes sharing one cache. The NVML snapshot it guards is node wide, so two processes with separate caches can probe at once and each can pick up the other's PID. @maverick-woo said the same on Project-HAMi/HAMi#2244.

The fix is a lock every container agrees on, but the only path they agree on is `/tmp/vgpulock`, which anyone on the node can write.

## Change

`flock` on a directory, so a killed owner doesn't strand waiters.

The wait takes an absolute deadline instead of a timeout, so a caller can spend one budget across this lock and the cache lock rather than give each a fresh 30s. Backoff goes 1ms to 100ms and the deadline is never renewed.

The path isn't trusted. Opening walks it a component at a time with `O_NOFOLLOW`, checks owner and mode at each level, and rejects filesystems where the lock wouldn't behave. It checks again after `flock` returns, since the path can be swapped while you're blocked.

One lock per process, `FD_CLOEXEC` so it doesn't cross an `exec`, and `after_fork()` drops the inherited descriptor so a child can't release a lock it never took.

An operation guard protects descriptor ownership during acquire and release. Overlapping calls return `EBUSY`; acquiring an already held lock returns `EDEADLK`. Once acquire returns, another thread in the same process may release it. This prevents an older acquisition's cleanup from closing a descriptor reused by a newer acquisition.

Operations defer thread cancellation until temporary descriptors have been closed or the acquired descriptor is owned by a cleanup handler. A cancelled acquisition releases its descriptor and clears the guard. The fork child clears the inherited guard along with its descriptor. Asynchronous cancellation is unsupported.

Nothing calls it yet. The call site needs `lock_postinit_deadline()`, which isn't on main, so it goes in with the integration.

## Test

CTest, no GPU and no driver. Covers acquire and release, the deadline API, a deadline that must not be renewed, a live holder timing out, a cancelled waiter, owner death, fork and exec cleanup, an untrusted path, and the path or an ancestor changing while a waiter is blocked. Cases that need `/proc` or Linux mount behaviour skip themselves and say so.

`test_hostpid_fallback_lock_threads` covers overlapping acquire/release, cancellation cleanup and a fork child inheriting a busy guard. It uses real threads and file locks with the production implementation. The Docker build runs both fallback tests with CTest.

```
cmake --build build --target test_hostpid_fallback_lock test_hostpid_fallback_lock_threads
ctest --test-dir build -R hostpid_fallback_lock
```

---
I used AI assistance (deepseek v4 flash) for the original split and Codex for these fixes. The lock design comes from #251.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -14,7 +14,7 @@ add_subdirectory(cuda)
 add_subdirectory(nvml)
 
 set(LIBVGPU vgpu)
-add_library(${LIBVGPU} SHARED libvgpu.c utils.c log_utils.c $<TARGET_OBJECTS:nvml_mod> $<TARGET_OBJECTS:cuda_mod> $<TARGET_OBJECTS:allocator_mod> $<TARGET_OBJECTS:multiprocess_mod>)
+add_library(${LIBVGPU} SHARED libvgpu.c utils.c log_utils.c hostpid_fallback_lock.c $<TARGET_OBJECTS:nvml_mod> $<TARGET_OBJECTS:cuda_mod> $<TARGET_OBJECTS:allocator_mod> $<TARGET_OBJECTS:multiprocess_mod>)
 target_compile_options(${LIBVGPU} PUBLIC ${LIBRARY_COMPILE_FLAGS})
 target_link_libraries(${LIBVGPU} PUBLIC -lcuda -lnvidia-ml)
 
--- a/src/hostpid_fallback_lock.c
+++ b/src/hostpid_fallback_lock.c
@@ -249,6 +249,18 @@ static int open_directory_without_symlinks(const char *path,
 }
 
 #ifdef __linux__
+/*
+ * ZFS is out of tree, so <linux/magic.h> never defines its magic and the
+ * check below would be compiled away.  OVERLAYFS_SUPER_MAGIC only arrived in
+ * 4.6, so an older build host drops overlay support the same way.
+ */
+#ifndef OVERLAYFS_SUPER_MAGIC
+#define OVERLAYFS_SUPER_MAGIC 0x794c7630
+#endif
+#ifndef ZFS_SUPER_MAGIC
+#define ZFS_SUPER_MAGIC 0x2fc12fc1
+#endif
+
 static int filesystem_type_supported(uint64_t filesystem_type) {
 #ifdef EXT4_SUPER_MAGIC
     if (filesystem_type == (uint64_t)EXT4_SUPER_MAGIC) {
--- a/src/hostpid_fallback_lock.c
+++ b/src/hostpid_fallback_lock.c
@@ -528,6 +528,10 @@ static int acquire_at_until(const char *path, uid_t trusted_owner,
     return 0;
 }
 
+#ifdef HOSTPID_FALLBACK_LOCK_TESTING
+/* Tests point the lock at a fixture directory.  Nothing else does, and the
+ * fixtures live under /tmp, whose ancestors are not owned by the test user,
+ * so the ancestor checks stay off here. */
 int hostpid_fallback_lock_acquire_at(const char *path, uid_t trusted_owner,
                                      unsigned int timeout_ms) {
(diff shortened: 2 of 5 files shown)


src/hostpid_fallback_lock.c
```diff
@@ -0,0 +1,584 @@
...
+static int validate_readonly_filesystem(int fd) {
+    struct statvfs filesystem_stat;
+
+    if (fstatvfs(fd, &filesystem_stat) != 0) {
+        return -1;
+    }
+    if ((filesystem_stat.f_flag & ST_RDONLY) == 0) {
+        errno = EACCES;
+        return -1;
+    }
```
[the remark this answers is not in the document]
iemAnshuman: I don't think they're conflated. the O_NOFOLLOW walk in open_directory_without_symlinks checks every component either way, require_readonly doesn't gate it. the flag only gets used in one spot, the ST_RDONLY check and that is on for the default path and off for acquire_at since a test dir under /tmp isn't read only

src/CMakeLists.txt
```diff
@@ -14,9 +14,9 @@ add_subdirectory(cuda)
 add_subdirectory(nvml)
 
 set(LIBVGPU vgpu)
-add_library(${LIBVGPU} SHARED libvgpu.c utils.c log_utils.c $<TARGET_OBJECTS:nvml_mod> $<TARGET_OBJECTS:cuda_mod> $<TARGET_OBJECTS:allocator_mod> $<TARGET_OBJECTS:multiprocess_mod>)
+add_library(${LIBVGPU} SHARED libvgpu.c utils.c log_utils.c hostpid_fallback_lock.c $<TARGET_OBJECTS:nvml_mod> $<TARGET_OBJECTS:cuda_mod> $<TARGET_OBJECTS:allocator_mod> $<TARGET_OBJECTS:multiprocess_mod>)
```
mesutoezdil, line 17: nothing calls this yet and the caller needs lock_postinit_deadline. what happens to these 676 lines if that follow-up stalls?

iemAnshuman: it does not get built into libvgpu.so right now when nothing call it. so if the follow up stalls it would just sit there. taking it out of the vgpu target so only the tests build it

iemAnshuman: the caller need lock_postinit_deadline() and broker client and neither is on main right now. which is why the caller isnt here. would you rather that I keep this as a standalone lock pr or hold it until the integration is ready?

iemAnshuman: done in the latest push

test/test_hostpid_fallback_lock.c
```diff
@@ -0,0 +1,903 @@
...
+
+int main(int argc, char **argv) {
+    char directory[] = "/tmp/hami-hostpid-global-lock.XXXXXX";
+    char *path;
+
+    if (argc == 3 && strcmp(argv[1], "--check-closed") == 0) {
+        int fd = atoi(argv[2]);
+
+        errno = 0;
+        return fcntl(fd, F_GETFD) == -1 && errno == EBADF ? 0 : 2;
```
mesutoezdil: fd numbers get reused after exec. is EBADF guaranteed?

mesutoezdil:
test_hostpid_fallback_lock_threads is not in the diff. dropped?

test/test_hostpid_fallback_lock.c
iemAnshuman: no but reuse only makes the test fail. will change it to compare dev/ino of the lock dir

iemAnshuman:
> test_hostpid_fallback_lock_threads is not in the diff. dropped?

yes. since operation guard and cancellation handling dropped too which it tested.

PR #317, opened 2026-09-04
fix: make SM limiter step linear in SM count

### LINKED ISSUE #314: CUDA_DEVICE_SM_LIMIT does not separate tenants sharing a GPU: utilization watcher correction step scales quadratically with SM count
First, English is not my native language, so I used AI to help with the translation. However, I performed all measurements and analysis myself, and the conclusions are my own.

## Previous discussion

I documented the behavior seen with a single process in #274 and submitted a fix in #286. Both issues were closed on August 21.

At that time, I only tested a single capped process while the GPU was otherwise idle. The additional tests here also cover the case where two processes share the same GPU.

## Environment

RTX PRO 6000 Blackwell Server Edition  
`sm_num=188`, `max_threads_per_sm=1536`  
pool `9,240,576`, `FACTOR=32`  
Driver 580.173.02, CUDA 13.3

Measurements were taken against HAMi-core commit `5496322`.

I used `GPU_CORE_UTILIZATION_POLICY=FORCE` with standalone `LD_PRELOAD`, and assigned a separate shared cache to each process.

## Symptoms

I ran two processes on a single GPU, with caps of 30 and 70 respectively. There were no other workloads on the GPU. Each 4096 FP32 matmul test ran for 240 seconds.

The results were as follows:

```text
                    A(30)      B(70)       Total   device util
No cap               23.1 it/s  23.1 it/s   46.2    98%
main                 21.4       25.2       46.6    98%
patched              13.2       34.9       48.1    98%
```

With the patched version, the caps are applied as intended without reducing the total throughput.

For comparison, on the same hardware, running a single process with a cap of 30 gives `main` a performance of 62.6% relative to the uncapped result. With the patched version, performance drops to 29.9%.

The difference between the two versions becomes larger as the cap is lowered. On the same 188-SM device, running a single process with a cap of 10 caused `main`’s `userutil` to continuously fluctuate between 0 and 83. The average `device util` was 26.9%, more than twice the target.

The patched version converged after approximately 20 seconds. Its `userutil` stayed between 9 and 13, with an average `device util` of 10.9%.

The acceleration branch multiplies by `diff * 2 / (up_limit + 1)`. As the cap gets lower, `diff` becomes larger, increasing the impact of this branch.

The cap-10 results were measured while working on #286.

## Root cause

`delta()` calculates the adjustment range as proportional to `sm_num * sm_num`.

```c
int64_t increment =
    (int64_t)g_sm_num[device_id] * (int64_t)g_sm_num[device_id] *
    (int64_t)g_max_thread_per_sm[device_id] * (int64_t)utilization_diff / 2560;
```

However, the size of the token pool scales linearly with the number of SMs:

```text
g_total_cuda_cores = sm_num * max_thread_per_sm * FACTOR
```

As a result, the fraction of the pool consumed by one tick is:

```text
sm_num * diff / 81920
```

In other words, the amount consumed per tick grows with the size of the device. On a 188-SM device, the pool is refilled within a single 120 ms tick. The limiter therefore fails to converge and repeatedly oscillates between saturation and throttling.

This causes the time-averaged utilization to remain significantly above the configured value.

If one `sm_num` factor is removed, the ratio becomes:

```text
diff / 81920
```

This makes the behavior independent of the device’s SM count.

## Logs

I have attached two archives

[sm-cotenant-evidence.zip](https://github.com/user-attachments/files/31629109/sm-cotenant-evidence.zip) — the two-process contention runs described above:

Watcher logs (LIBCUDA_LOG_LEVEL=3, one line every 120 ms)
nvidia-smi samples
Iteration results recorded at 10-second intervals
The execution script and the load program

[sm-limit10-188sm-evidence.zip](https://github.com/user-attachments/files/31629113/sm-limit10-188sm-evidence.zip) — the cap-10 runs, measured while working on [#286](https://github.com/MayMustAI/platform-core/issues/286):

Watcher logs and nvidia-smi samples for main and the patched build

Both libraries were built from the same source tree using the same toolchain.

maymust-jonghyunlee:
First, English is not my native language, so I used AI to help with the translation. 

Fixes #314

Removes one `sm_num` factor from the correction step in `delta()`.
The token pool is:
```
sm_num * max_thread_per_sm * FACTOR
```
The correction step previously contained:
```
sm_num * sm_num * max_thread_per_sm * diff / 2560
```
This caused the step-to-pool ratio to increase with the SM count. The ratio is now diff / 81920 regardless of device size.
This PR does not change g_total_cuda_cores in setspec(). Its int64_t conversion is covered by #303, and the two changes can be merged in either order.

On a 188-SM GPU, the previous correction step could refill the entire pool in a single 120 ms tick. This caused the limiter to alternate between saturation and throttling instead of settling at the configured cap. See #314 for logs and further analysis.

Test environment:
- RTX PRO 6000 Blackwell Server Edition (188 SMs)
- Driver 580.173.02
- CUDA 13.3
- Standalone LD_PRELOAD
- GPU_CORE_UTILIZATION_POLICY=FORCE
- Separate shared cache per process
Two processes shared one GPU with caps of 30 and 70, running 4096 FP32 matrix multiplication for 240 seconds:

| build | A(30) it/s | B(70) it/s | total | device util |
|---|---|---|---|---|
| no cap | 23.1 | 23.1 | 46.2 | 98% |
| main | 21.4 | 25.2 | 46.6 | 98% |
| this change | 13.2 | 34.9 | 48.1 | 98% |

With one process capped at 10, main averaged 26.9% device utilization while userutil oscillated between 0 and 83. With this change, userutil converged within approximately 20 seconds and remained between 9 and 13, averaging 10.9%.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected CUDA utilization calculations to prevent overestimating core activity on devices with multiple streaming multiprocessors.

### DIFF
--- a/src/multiprocess/multiprocess_utilization_watcher.c
+++ b/src/multiprocess/multiprocess_utilization_watcher.c
@@ -87,7 +87,7 @@ static int64_t delta(int up_limit, int user_current, int64_t share, int device_i
   int utilization_diff =
       abs(up_limit - user_current) < 5 ? 5 : abs(up_limit - user_current);
   int64_t increment =
-      (int64_t)g_sm_num[device_id] * (int64_t)g_sm_num[device_id] *
+      (int64_t)g_sm_num[device_id] *
       (int64_t)g_max_thread_per_sm[device_id] * (int64_t)utilization_diff / 2560;
 
   /* Accelerate cuda cores allocation when utilization vary widely */


maymust-jonghyunlee:
Not like #286 LLM helped posting my replies.
This time i used LLM for translating description.

src/multiprocess/multiprocess_utilization_watcher.c
```diff
@@ -87,7 +87,7 @@ static int64_t delta(int up_limit, int user_current, int64_t share, int device_i
   int utilization_diff =
       abs(up_limit - user_current) < 5 ? 5 : abs(up_limit - user_current);
   int64_t increment =
-      (int64_t)g_sm_num[device_id] * (int64_t)g_sm_num[device_id] *
+      (int64_t)g_sm_num[device_id] *
```
mesutoezdil, line 90: this also shrinks the step by sm_num on small cards. did convergence still hold on something like a 40-sm gpu?

maymust-jonghyunlee: Since I don't have 40 sm gpu (maybe like 4080?) i tested with gtx 1080.

Measurements were taken on a gtx 1080 with 20 sm (Pascal, max_threads_per_sm=2048, pool size 1,310,720), using driver 580.173.02,CUDA_DEVICE_SM_LIMIT=50,GPU_CORE_UTILIZATION_POLICY=FORCE. Ran continuous matmul for 240 seconds using standalone LD_PRELOAD. Both were built from the same source tree with the same toolchain: main versus main with this patch applied. Watcher samples were collected approximately every 20 seconds.

On main, userutil was 4**6, 62, 9, 100, 17, 83, 95, 100, 100, 100, 100, 47,** unstable. Later in the run, it stayed at 100 for roughly 80 seconds, while share alternated between 0 and approximately 95000. Average device utilization, sampled every 5 seconds with nvidia-smi, was 81%, well above the 50% target.

With the patch, userutil was **40, 41, 43, 55, 56, 68, 51, 67, 50, 43, 42.** It was already within the target band at the first sample, 20 seconds into the run, and share remained stable between 95000 and 102000 Average device utilization was 49%, close to the 50% target.

mesutoezdil: This is being closed because it does not comply with the contribution guidelines. Point 4.

Issue #233, opened 2026-07-27
Compatible with the CUDA Driver API library functions for CUDA 13.x


xxsoul:
CN：阅读代码时我发现HAMi-Core虽然基于CUDA 13.x 版本的镜像进行编译，但是还没有增加对 CUDA 13.x 版本新引入的函数劫持。在上层应用使用CUDA 13.x 版本提供的新函数时，可能会报错找不到符号等问题。

EN：Whilst reading through the code, I noticed that although HAMi-Core is compiled using a CUDA 13.x image, it does not yet support the function hijacking feature newly introduced in CUDA 13.x. When upper-layer applications use the new functions provided by CUDA 13.x, errors such as ‘symbol not found’ may occur.

xxsoul:
CN：我准备本周开始着手增加对新函数的劫持工作，尽量识别其中与显存分配相关的函数，按已有风格增加对显存的限制（如果可能）。

EN：I plan to start working on intercepting new functions this week, identifying as many as possible that relate to video memory allocation, and adding restrictions on video memory in line with the existing approach (where possible).

maverick-woo:
Do you have any error logs or specific missing symbols? In theory, HAMi only needs to hook compute and memory-related APIs. Most other APIs can pass through, although some memory-copy APIs may need special handling. Please provide the exact APIs and a reproducer.

badnikhil:
@maverick-woo Here is the audit you asked for. Short version: **I could not reproduce a
"symbol not found" on CUDA 13.x, and I do not think it can happen** — but the audit did turn up one
real 13.x regression, in `cuCtxCreate`.

## 1. Where the symbol list comes from

Not from the docs site — from the ABI NVIDIA ships. Each CUDA release's redistributable archive
contains `cuda_cudart/lib/stubs/libcuda.so`, which is the exact link-time driver surface for that
release:

```bash
BASE=https://developer.download.nvidia.com/compute/cuda/redist
curl -sO $BASE/cuda_cudart/linux-x86_64/cuda_cudart-linux-x86_64-13.3.29-archive.tar.xz
tar -xf cuda_cudart-linux-x86_64-13.3.29-archive.tar.xz
nm -D --defined-only cuda_cudart-*/lib/stubs/libcuda.so | awk '$3 ~ /^cu/ {print $3}' | sort -u
```

[...]

PR #320, opened 2026-09-09
fix: map cuCtxCreate to the right version on CUDA 12.x and 13.x

badnikhil:
Refs #233.

`find_symbols_in_table()` probes `_v3`, `_v2`, then the bare name. `cuda.h` remaps `cuCtxCreate` to `_v4` on 13.x, so the ladder can never reach it, and the `_v2`/`_v3` hooks are compiled out under `#if CUDA_VERSION < 13000`. The hook is dead on every 13.x build. On 12.x the probe hits `_v3`, whose signature does not match the `cuCtxCreate` the caller holds.

Two rows in `g_func_map`, which exists for this.

| build | caller cudaVersion | before | after |
|---|---|---|---|
| 12.9 | 12090 | `_v3` (wrong signature) | `_v2` |
| 13.3 | 13030 | pass-through | `_v4` |
| 13.3 | 12090 | pass-through | pass-through |

GPU-free regression test in the `test_nvml_v2_signatures` style, fails 3 of 4 cases without the fix. `hack/check_cuda_hook_consistency.py` passes unchanged.

Not adding the 53 new 13.x pass-through symbols: they already pass through, and `cuCtxGetDevice_v2`/`cuCtxSynchronize_v2` would break the base names since the `_v2` probe runs first.

@xxsoul offered to test on NVIDIA 610 driver nodes.

AI-assisted. I reviewed every line of this diff before submitting.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CUDA compatibility for context creation across CUDA versions 10–13 and newer.
  - Applications using different CUDA runtime versions now resolve the appropriate context creation interface more reliably.
  - Added fallback handling when a compatible symbol is unavailable, allowing calls to pass through to the driver.

- **Tests**
  - Added GPU-free regression coverage for CUDA version combinations, including compatibility and fallback scenarios.


src/include/multi_func_hook.h
```diff
@@ -25,7 +25,15 @@ static CudaFuncMapEntry g_func_map[] = {
...
     {"cuGraphKernelNodeSetParams", 10000, 11999, "cuGraphKernelNodeSetParams"},
-    {"cuGraphKernelNodeSetParams", 12000, 99999, "cuGraphKernelNodeSetParams_v2"}
+    {"cuGraphKernelNodeSetParams", 12000, 99999, "cuGraphKernelNodeSetParams_v2"},
+
+    // cuda.h remaps cuCtxCreate to cuCtxCreate_v2 up to CUDA 12.x and to
+    // cuCtxCreate_v4 from CUDA 13.0.  find_symbols_in_table() only probes the
+    // _v3, _v2 and bare spellings, so it can never reach the _v4 hook, and on
+    // 12.x it answers with _v3, whose five-parameter signature does not match
+    // the three-parameter cuCtxCreate the caller has.  Pin both explicitly.
+    {"cuCtxCreate", 10000, 12999, "cuCtxCreate_v2"},
```
mesutoezdil, line 35: the shipped image builds against cuda 13.3, where cuCtxCreate_v2 is compiled out. can this row ever resolve outside the test stub?

xxsoul:
# Testing Process:
My testing procedure was as follows: I compiled the PR version and the 2.10 version of libvgpu.so (libvgpu-pr320 and libvgpu-main.so) on the nvidia/cuda:13.3.0-devel-ubi8 image, then wrote a simple CUDA application to simulate a workload and call the `cuCtxCreate` function, and finally built the corresponding binary applications (pr320_ctx_workload-cu13, pr320_ctx_workload-cu12), and then launched the applications on a node running driver version 610.57.04 (CUDA 13.3) using the `nvidia/cuda:13.3.0-devel-ubi8` image.

# Test Results
## Regarding libvgpu.so
- In this example, only pr320_ctx_workload-cu13 was used for testing, yielding the following results:
1. libvgpu-main.so only intercepted direct symbol usage; it did not intercept the cuCtxCreate function pointer obtained via cuGetProcAddress: image
2. Both the direct symbol call in libvgpu-pr320.so and the `cuCtxCreate` function obtained via `cuGetProcAddress` were hijacked: image
3. Based on this use case, I believe the current change is correct and valid @badnikhil 

## Regarding pr320_ctx_workload-cu12
- I saw @mesutoezdil’s discussion, so I added a simulated workload application compiled with CUDA 12.9.1. I tested it using both pr320_ctx_workload-cu13 and pr320_ctx_workload-cu12 in the libvgpu-pr320 environment and obtained the following results
1. pr320_ctx_workload-cu13: The application runs normally and successfully calls the `cuCtxCreate` function in `libvgpu.so` : image
2. pr320_ctx_workload-cu12: The application runs normally, but the `cuCtxCreate` symbol in `libvgpu.so` is not hit; instead, the fallback hits `/usr/lib/x86_64-linux-gnu/libcuda.so.1` : image
3. **I'm not sure if this behavior is normal?**

P.S.: [pr320_ctx_workload.c](https://github.com/user-attachments/files/32431998/pr320_ctx_workload.c) is the source code for the CUDA application I used during testing. Please let me know if you spot any issues.

archlitchi:
please resolve this conflicts

src/include/multi_func_hook.h
```diff
@@ -25,7 +25,15 @@ static CudaFuncMapEntry g_func_map[] = {
...
+    {"cuCtxCreate", 13000, 99999, "cuCtxCreate_v4"}
```
mesutoezdil: on 13.x cuCtxCreate now hits the table, and _cuGetProcAddress_v2 returns success without writing *symbolStatus, while cuGetProcAddress_v2 at hook.c:409 does. is the caller fine reading it uninitialised?

src/include/multi_func_hook.h
mesutoezdil: check_cuda_hook_consistency.py never reads multi_func_hook.h, so a real_name that is misspelled or compiled out here is not caught. worth extending it to g_func_map?

mesutoezdil:
#233 asks for hooks on the functions CUDA 13.x newly introduced, and this adds none, so the issue should stay open after this merges. the conflict is only in test/CMakeLists.txt, multi_func_hook.h is untouched on main since 9 Sep, so a rebase should be mechanical.

mesutoezdil:
@archlitchi conflicts are gone, rebased and squashed to one commit.

@xxsoul the cu12 fallback you saw is expected. A 13.3 build has no `cuCtxCreate_v2` wrapper, so there is nothing for that spelling to hit. Thanks for testing on real hardware.

Two more symbols had the same bug, fixed here too. Details in the description.

PR #322, opened 2026-09-11
fix(cuda): handle null meminfo outputs and free untracked pointers

### LINKED ISSUE #321: [Bug] OptiX/Vulkan-CUDA interop: NULL cuMemGetInfo output crashes and external mapped buffers fail cuMemFree
### Summary

We reproduced two consecutive HAMi-core compatibility failures in a CUDA/Vulkan renderer using the NVIDIA OptiX denoiser:

1. OptiX calls `cuMemGetInfo_v2(NULL, &total)`. The intercepted function dereferences the NULL output pointer and crashes.
2. After guarding that query in an isolated diagnostic process, releasing CUDA buffers mapped from Vulkan external memory fails: `cuMemFree_v2 → free_raw → remove_chunk` returns `-1` for an untracked pointer instead of freeing it through the driver.

The NULL-query failure is related to #234. This report adds an argument-level reproduction and a second, independently verified external-memory release failure that remains after the first is bypassed.

### Environment

- Linux x86_64
- NVIDIA driver: `580.126.09`; `cuDriverGetVersion` returns `13000`
- HAMi-core injected through the upstream KAI-resource-isolator chart `1.1.0-chart`, using `/etc/ld.so.preload`
- Exact bundled HAMi-core commit has not yet been determined. The function behavior below was checked against the installed library's debug information and disassembly, not assumed from current main.
- Workload: an application renderer with OptiX denoising and Vulkan/CUDA external-memory interoperability

### 1. NULL output pointer in cuMemGetInfo_v2

A diagnostic interposer captured the following simplified stack and arguments before forwarding to the original HAMi function:

```text
renderer denoiser initialization
  → libnvoptix.so.1
  → cuMemGetInfo_v2(free=NULL, total=<valid pointer>)
  → libvgpu.so
  → SIGSEGV on an output-pointer write
```

In the installed binary, the faulting instruction resolves to `src/cuda/memory.c:508` inside `cuMemGetInfo_v2`.

A standalone Python probe, run separately for each case:

```python
import ctypes as C
import resource
import sys

resource.setrlimit(resource.RLIMIT_CORE, (0, 0))
cuda = C.CDLL("libcuda.so.1")

def check(name, *args):
    rc = getattr(cuda, name)(*args)
    if rc != 0:
        raise RuntimeError(name, rc)

check("cuInit", 0)
device = C.c_int()
check("cuDeviceGet", C.byref(device), 0)
context = C.c_void_p()
check("cuDevicePrimaryCtxRetain", C.byref(context), device)
check("cuCtxSetCurrent", context)

case = sys.argv[1]
free = C.c_size_t()
total = C.c_size_t()
free_arg = None if case in ("free-null", "both-null") else C.byref(free)
total_arg = None if case in ("total-null", "both-null") else C.byref(total)
print("calling", case, flush=True)
print("return code:", cuda.cuMemGetInfo_v2(free_arg, total_arg), flush=True)
check("cuDevicePrimaryCtxRelease_v2", device)
```

Run with `valid`, `free-null`, `total-null`, and `both-null`, each in its own process. In the injected environment, verify that the resolved query points into `libvgpu.so`; the clean control must actually have no HAMi injection, including no system preload.

| Arguments | Native NVIDIA driver | HAMi interception |
|---|---|---|
| Both outputs non-NULL | CUDA_SUCCESS | CUDA_SUCCESS |
| free=NULL | CUDA_SUCCESS | SIGSEGV |
| total=NULL | CUDA_SUCCESS | SIGSEGV |
| Both NULL | CUDA_SUCCESS | SIGSEGV |

The driver query controls ran on the same host. The public `cuMemGetInfo` documentation does not explicitly promise NULL-output support, so this is reported as incompatibility with observed NVIDIA driver/OptiX behavior, rather than a claimed documented guarantee for every driver version. The NULL argument in the renderer originates inside NVIDIA OptiX.

### 2. External mapped buffers cannot be freed

The renderer follows this lifecycle:

```text
Vulkan memory allocation/export
  → cudaImportExternalMemory
  → cudaExternalMemoryGetMappedBuffer
  → use mapped buffer
  → cudaFree(mapped_buffer)
  → cudaDestroyExternalMemory
```

NVIDIA explicitly requires the pointer returned by `cuExternalMemoryGetMappedBuffer` to be released using `cuMemFree`:
[CUDA external resource interoperability documentation](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXTRES__INTEROP.html).

In the installed HAMi library:

- `cuExternalMemoryGetMappedBuffer` forwards to the driver without registering the pointer in the allocation list.
- `cuMemFree_v2` calls `free_raw`, which calls `remove_chunk(device_overallocated, ptr)`.
- When the pointer is absent, `remove_chunk` returns `-1` without calling the real driver free.

The captured failure stack is:

```text
renderer external-buffer cleanup
  → CUDA runtime cudaFree
  → HAMi cuMemFree_v2
  → free_raw
  → remove_chunk: pointer not found, return -1
```

The runtime reports an unrecognized error code. This is not evidence of an invalid application free: in the diagnostic test, passing each of these pointers to the actual driver free succeeded.

### Controlled rendering results

The minimal renderer test uses a generated cube, 64×64 output, and one camera view. No inference model is loaded. The renderer itself is not a public standalone reproducer; the independently runnable NULL-query probe is included above.

| Configuration | Result |
|---|---|
| No HAMi, dedicated GPU, identical image digest and renderer binary, same driver version | Single-view render completed |
| HAMi + diagnostic NULL-output guard only | Fails during mapped-buffer release; exit 134 |
| HAMi + NULL guard + real-driver free for untracked pointers | Render completed; exit 0 |

In the successful shared-GPU diagnostic run:
- Three ordinary CUDA allocations were freed by HAMi's normal path successfully.
- Four externally mapped buffer pointers returned `-1` from HAMi's original release path.
- Forwarding those four releases to the actual driver returned `CUDA_SUCCESS` for every pointer.

The rendering control used a different host with the same driver version. The diagnostic changes were process-local interposition only; this does not constitute validation of a production patch or the broader memory-accounting semantics.

### Expected behavior / possible fix direction

- Preserve compatible handling of NULL output pointers in the memory-info hook while retaining the configured virtual-memory limit and propagating driver errors. Using local output variables for the driver query and copying only to non-NULL caller pointers is one possible approach.
- Correctly release valid externally mapped pointers that are not in the normal allocation list, and propagate the real driver result. Simply returning success without actually freeing would leak mappings.
- Add regression coverage for NULL queries and the import → map → free → destroy lifecycle, including both tracked and untracked pointers.

Could maintainers confirm the intended external-memory tracking behavior and whether fixes for these paths are already in progress?

liuxu623:
OptiX can call `cuMemGetInfo_v2` with an omitted output, which currently causes the hook to dereference NULL. Separately, CUDA pointers returned by external-memory mapping are not in the allocator's tracked list, so a valid `cuMemFree_v2` call returns `-1` instead of reaching the driver.

This change queries memory information into local variables, propagates driver errors, and writes only requested outputs while preserving virtual memory reporting and clamping subtraction at zero. Synchronous frees of untracked pointers now reach the real driver and return its result without changing allocation accounting. The existing tracked-pointer path is unchanged.

Fixes #321.

### Validation

- Added a GPU-free regression test linked to the production memory wrappers and allocator, registered with CTest. Direct GCC builds passed with debug logging enabled; the new test also passes the repository's cpplint configuration.
- Replacing the fixed memory-query implementation with the original produces SIGSEGV. Replacing the fixed allocator implementation with the original produces eight failed assertions.
- Tested on real NVIDIA GPU hardware with driver 580.126.09 (CUDA Driver API 13000), using the same application image and GPU for all three paths. The device model is intentionally omitted.

| Path | Valid outputs | Either/both outputs NULL | Minimal OptiX/Vulkan-CUDA rendering |
| --- | --- | --- | --- |
| Previously deployed library | Pass | SIGSEGV | SIGSEGV |
| Library built from this change | Pass | Pass | Pass, exit 0 |
| Native NVIDIA driver | Pass | Pass | Pass, exit 0 |

The patched render completed all seven logged synchronous frees successfully. Library provenance was checked with `dladdr`; no diagnostic workaround shim was loaded. The full production library was compiled directly with GCC inside the target container using the hook/multiprocess options from `build.sh`; a complete CMake build was not run.

### Scope

The hardware comparison used a dedicated GPU with per-process library loading and a configured virtual memory quota to avoid loading both library versions together. Virtual memory reporting was verified, but shared scheduling, concurrent isolation, and full application end-to-end behavior were not validated.

This fixes compatibility and release behavior; it does **not** add quota enforcement or allocation tracking for Vulkan-owned external memory. Reporting a virtual quota does not demonstrate that external allocations are constrained by it. Existing error/accounting behavior for tracked frees is outside this change.

AI assistance was used for implementation, tests, and this description.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved GPU memory reporting when requesting only selected values.
  - Corrected memory totals when usage reaches or exceeds configured limits.
  - Ensured externally allocated or untracked memory is properly released.
  - Improved handling of invalid contexts and driver-reported errors during memory operations.

- **Tests**
  - Added regression coverage for GPU memory queries and release behavior, including partial outputs, error handling, accounting updates, and lock safety.


src/allocator/allocator.c
```diff
@@ -206,7 +202,9 @@ int remove_chunk(allocated_list *a_list, CUdeviceptr dptr) {
     }
 
     pthread_mutex_unlock(&mutex);
-    return -1;
+    /* External-memory mappings and other untracked pointers still belong
+     * to the driver. Do not adjust accounting for memory we did not track. */
+    return cuMemoryFree(dptr);
```
mesutoezdil, line 207: #258 has been forwarding untracked pointers here since august. does this part need to land twice?

liuxu623: #258 is still open and `main` still returns `-1` here, so nothing has landed twice. This hunk is required for #321: OptiX external-memory mappings never go through `add_chunk`, so `cuMemFree_v2` on them fails without it. Happy to rebase and drop it if #258 merges first.

mesutoezdil: #258 merged since, 410dfbe. main forwards here now.

mesutoezdil:
resolve conflicts pls

src/allocator/allocator.c
mesutoezdil: #342 landed the rest. keep the test, drop the code.

src/cuda/memory.c
```diff
@@ -490,28 +490,27 @@ CUresult cuMemGetInfo_v2(size_t* free, size_t* total) {
     CHECK_DRV_API(cuCtxGetDevice(&dev));
     size_t usage = get_current_device_memory_usage(cuda_to_nvml_map(dev));
     size_t limit = get_current_device_memory_limit(cuda_to_nvml_map(dev));
```
mesutoezdil: main maps only the usage here now. this would revert #340.

PR #326, opened 2026-09-12
fix: preserve CUDA dlsym caller handle

changluyi:
/kind bug

## What type of PR is this?

Fixes a runtime crash in the CUDA dlsym hook path.

## What this PR does / why we need it

When no CUDA redirect library is configured, the `dlsym` hook previously ignored the caller-provided handle for CUDA symbols and always queried the redirect handle. If that handle was unavailable, this became an equivalent to a global lookup. On WSL2, the CUDA driver performs an internal `dlsym(handle, \"cuInit\")`; the global lookup resolved back to HAMi's own `cuInit` hook and recursed until the process crashed with SIGSEGV.

This change preserves the caller's handle when no redirect library was loaded. It also extends the dlsym regression test to resolve `cuInit` from `libcuda.so.1` and verify that the result does not come from `libvgpu.so`.

## Which issue(s) this PR fixes

None.

## Validation

- WSL2
- NVIDIA GeForce RTX 5060 Laptop GPU
- NVIDIA driver 591.86
- CUDA 13.3 container: `docker-mirrors.alauda.cn/nvidia/cuda:13.3.0-cudnn-devel-ubi8`
- `test_dlsym_rtld_next` without preload: passed
- `test_dlsym_rtld_next` with `LD_PRELOAD=build/libvgpu.so`: passed
- `test_mem_create`: passed (4/4 threads)
- `test_runtime_alloc`: passed
- `python3 hack/check_cuda_hook_consistency.py`: passed

The full `make build-in-docker` rerun was blocked by an intermittent package repository timeout while installing build dependencies; the affected library and regression test were rebuilt and exercised in the CUDA container.

## User-facing change

No API change. CUDA driver internal symbol lookup no longer recurses through the HAMi preload hook when no redirect library is configured.

## AI disclosure

AI assistance was used for repository exploration, debugging support, and draft wording. The code change, test design, and validation results were reviewed by the author.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected CUDA symbol lookup when no redirect library is loaded, preserving the caller-provided handle.
  * Ensured CUDA driver functions resolve from the intended library instead of the preload library.

* **Tests**
  * Added coverage verifying successful resolution of CUDA driver entry points through dynamic symbol lookup.


changluyi:
Updated in commit caf4f25: concrete caller handles now always take precedence over CUDA_REDIRECT, preventing CUDA driver internal lookups from being redirected back into HAMi. The regression test was also run with CUDA_REDIRECT pointing to libvgpu.so and passed.

PR #330, opened 2026-09-14
fix: compute total cuda cores in nvmlDeviceGetHandleByIndex.

### LINKED ISSUE #318: SM utilization is read from the wrong GPU when CUDA_VISIBLE_DEVICES is not identity
`get_used_gpu_utilization()` in `src/multiprocess/multiprocess_utilization_watcher.c:221` maps the loop index to a CUDA index, then passes that CUDA index to `nvmlDeviceGetHandleByIndex()`, which wants an NVML index.

Only correct when the map is identity. With `CUDA_VISIBLE_DEVICES=2,3` the watcher samples NVML 0 and 1, GPUs the container was never given, and stores their utilization as its own. `CUDA_DEVICE_SM_LIMIT` is then driven by another tenant's load.

`set_task_pid()` in `src/utils.c` runs the same loop and passes `i`, not the mapped value.

Repro without a GPU, stub NVML in the style of `test/test_nvml_v2_signatures.c`:

```
CUDA 0 -> NVML 2, CUDA 1 -> NVML 3
userutil[CUDA 0] = 10   expected 30
userutil[CUDA 1] = 20   expected 40
```

Passing `devi` instead of `cudadev` gives 30 and 40. Happy to send the fix plus the harness as a GPU-free test.

maishivamhoo123:
Fixes #318

The functions like get_used_gpu_utilization() gives nvmlDeviceGetHandleByIndex() the CUDA index instead of the NVML index pass devi instead of cudadev. and also adds a GPUfree ctest that fails on main and passes with the fix.

Disclosure: test drafted with LLM and took the help of Claude while testing and drafting the code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected GPU utilization reporting when CUDA device mappings differ from NVML device indexes.
  * Ensured utilization, per-process activity, and monitored memory are attributed to the correct physical GPU.

* **Tests**
  * Added regression coverage that runs without GPU hardware.
  * Validated GPU subsets, single-device visibility, reversed device ordering, and identity mappings to help prevent future indexing errors.


src/multiprocess/multiprocess_utilization_watcher.c
```diff
@@ -219,7 +219,7 @@ int get_used_gpu_utilization(int *userutil,int *sysprocnum) {
         continue;
       userutil[cudadev] = 0;
       nvmlDevice_t device;
-      CHECK_NVML_API(nvmlDeviceGetHandleByIndex(cudadev, &device));
+      CHECK_NVML_API(nvmlDeviceGetHandleByIndex(devi, &device));
```
mesutoezdil, line 222: CUDA_VISIBLE_DEVICES=1 on a 4-gpu node leaves identity entries behind, so nvml 2 still maps to cudadev 2. does a case cover that?

mesutoezdil (commented):
the title looks like two titles pasted together, one of them from #303.

src/multiprocess/multiprocess_utilization_watcher.c
maishivamhoo123: Yes @mesutoezdil, you are right CUDA_VISIBLE_DEVICES=1 on a 4-GPU node leaves identity entries behind, and NVML 0 does not appear in the map, But this PR still works correctly for the visible device and does not touch parse_cuda_visible_env() or nvml_to_cuda_map(), so I think it should we should handle that separately in different issue. 
In my latest push, I added a test case for this scenario, which will verifie the CUDA 0 maps to NVML 1 instead of NVML 0, so the index 0 is never requested, and indexes 2 and 3 are still requested.

archlitchi (approved):
/lgtm

PR #331, opened 2026-09-15
fix(security-insights): make the insights file parseable and unset the image exec bit

mesutoezdil:
`security-insights.yml` targeted the unreleased 2.2.0 spec, where `in-scope` and `out-of-scope` are declared as string lists but generated as a single `*URL`, so `si.Load` rejected the whole file and every Baseline check reading it degraded, including OSPS-QA-04.01 reporting no repository list when both are present; it now declares 2.1.0, uses `security-policy` in place of `policy`, folds the scope text into the neighbouring comment, and passes both `cue vet` against the v2.1.0 schema and `si.Load`.

`docs/images/sample_nvidia-smi.png` was committed as mode 100755 and the scanner treats a binary blob carrying an execute bit as a generated executable artifact, which is why it tripped OSPS-QA-05.01 while `docs/images/hami-core-arch.png` at 100644 did not, so the mode is now 100644 with the content byte-identical.


archlitchi (approved):
/lgtm

PR #332, opened 2026-09-17
fix: load override env before shared-region init

Utkarsh56016:
## Summary

This PR is the HAMi-core/libvgpu runtime-side part of Project-HAMi/HAMi#2125.

It loads `/overrideEnv` before shared-region initialization so env-stripped login/SSH-started processes can recover their HAMi allocation config before libvgpu chooses the shared cache path and initializes memory/core limits.

This is intended to pair with a follow-up Project-HAMi/HAMi PR that writes and mounts the per-container `/overrideEnv` file from the device-plugin allocation path.

## Problem

For the #2125 reproduction, login-shell and SSH-started processes can start without the HAMi allocation environment inherited by the original container process.

In that path, libvgpu may still be loaded, but the allocation config can be unavailable before shared-region setup. When `CUDA_DEVICE_MEMORY_SHARED_CACHE` and `CUDA_DEVICE_MEMORY_LIMIT_*` are missing at initialization time, enforcement can fall back to the default shared cache path and the configured GPU memory limit is not applied consistently for the env-stripped process path.

## Change

This PR:

- calls `load_env_from_file(ENV_OVERRIDE_FILE)` at the start of `initialized()`, before shared-region/cache setup
- makes `ENV_OVERRIDE_FILE` test-overridable while keeping `/overrideEnv` as the production default
- adds a GPU-free regression test, `override_env_preinit`

The test verifies that memory limit, SM limit, and shared-region cache path are loaded from the override file before shared-region creation.

## Validation

Local validation on the rebased branch:

- `git diff --check upstream/main...HEAD`: PASS
- `cmake -S . -B /tmp/hami-libvgpu-override-pr-build -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -DCMAKE_BUILD_TYPE=RelWithDebInfo`: PASS
- `cmake --build /tmp/hami-libvgpu-override-pr-build --target test_override_env_preinit -j"$(nproc)"`: PASS
- `ctest --test-dir /tmp/hami-libvgpu-override-pr-build -R '^override_env_preinit$' --output-on-failure`: PASS

## Notes

This PR only adds the HAMi-core side of the fix. It does not create `/overrideEnv` by itself.

A follow-up dependent Project-HAMi/HAMi PR will provide the device-plugin side...

## Summary by CodeRabbit

- **New Features**
  - Override files are loaded before HAMi-core initialization, ensuring configured memory, SM, and shared-cache settings take effect consistently.
  - The override file location can be customized through the build environment.

- **Bug Fixes**
  - Corrected initialization so override settings are available before shared resources and process state are created.

- **Tests**
  - Added coverage confirming override values are applied and the configured shared-cache path is created successfully.


mesutoezdil:

